PR #2351, opened 2026-07-13
SuperPoint+LightGlue structural backend, 3-DoF geometry, and the screenshot-iconography study

samggreenberg:
## What

- **`vtscore/media/structural_geometry.py`** — parametrised geometric verification: the production 4-DoF similarity fit plus a new **3-DoF scale+translation model** (no rotation), with exhaustive 2-point minimal-sample RANSAC, emitting the exact production `MatchStats`/gates. Unit-tested (7 tests).
- **`vtscore/media/structural_splg.py`** — the **SuperPoint+LightGlue backend** the structural-embedder design doc reserves, StructuralMatcher-conformant, geometry pluggable.
- **`docs/reports/2026-07-13-screenshot-iconography.html`** — the full study these modules were built for (4 corpora, 463 queries, 1.6M matched pairs).

## Why (measured)

- On scanned-document corpora SP+LG **doubles SigLIP** (AP 0.395/0.481 vs 0.204/0.235) and verifies **41% of true pairs vs SIFT's 5%** — line-art stamps/logos were a SIFT problem, not a geometry problem.
- **3-DoF is a free precision win** on flat rasters: AP identical to 4-DoF, false verifications drop sharply (synth SIFT verified-precision 0.77→0.98). Recommend it as the default fit for screenshot/document media profiles.
- Below ~128 px targets, no correspondence backend works — small-icon search needs a different instrument (see report roadmap).

## Not yet wired

- Ingest integration needs an **fp16 descriptor path** in `StructuralFeatures.compact()` (uint8 cast destroys float SuperPoint descriptors — see module docstring).
- SP+LG must ship with its **own cold-gate inlier floor (~24, not 8)**: LightGlue's denser matches over-fire the SIFT-calibrated gate (5.7% pair FPR at floor 8 → 0.5% at 24, still 4× SIFT's TPR).
- Extra deps (not in default install): `pip install --no-deps git+https://github.com/cvg/LightGlue.git` + `kornia`. ("LightGlue v2" from the design doc does not exist publicly; this uses official cvg/LightGlue.)

Experiment artifacts + runnable pipeline: `/exp/sgreenberg/experiments/screenshot-search/` on the grid.


samggreenberg:
**Tiling extension added to the report** (follow-up question: is small-icon failure the small target, or the big canvas?)

Candidate-side 224 px / 50%-overlap tiling (mean 125 tiles/screenshot), whole-image vs tiles@1x vs tiles@2x on identical pair sets, plus max-over-tiles VLAD for Stage-1.

**It was mostly the canvas (keypoint-budget dilution), not the target:**
- Tiled **SIFT**: true-pair verify rate 0.2% → **14%** (32–64 px) and 3.8% → **20%** (64–128 px) at essentially unchanged FPR; replicates on real Rico icons (0.1% → 3.9%, 0.4% → 7.7%). No gate change needed.
- Tiled **SP+LG**: raw TPR 33–55% in every bucket **but 33% FPR** (125 tiles = 125 hypothesis tests for LightGlue to hallucinate on UI chrome). Gate-matched to FPR ≤ 1% (floor 24), only ≥ 64 px survives (128–256 px: 10.4% → 28%); tiled SIFT beats it below that. For small flat iconography the classic detector in a small window is the more trustworthy instrument.
- **Tile-VLAD Stage-1**: synth AP 0.020 → **0.110** (canonical-artwork queries → **0.178**, above even exhaustive direct SIFT search); Rico +33% AP, 3× R@50. Regional retrieval on the existing codebook — likely the highest-leverage production change in the whole study.
- The absolute size floor is real but narrow: only < 32 px stays broken (tile@2x SIFT recovers 8% there).

Roadmap item 3 in the report is updated accordingly: tile-VLAD Stage-1 + tiled SIFT Stage-2 is the small-icon path; only sub-32 px still needs a non-correspondence method.

PR #2628, opened 2026-07-18
Honest whole-job ETA for warm and uncalibrated dataset loads

### LINKED ISSUE #2615: Dashboard dataset-row ETA disagrees with the SSE eta_seconds
During the #2556 validation runs (2026-07-18), the dashboard loading row showed "2.5 min left?" during a slow decode phase while the SSE `loading-tasks` payload's `eta_seconds` was ~3,300 s for the same instant. The frontend appears to derive a phase-local ETA from `current`/`total` plus its own timing when counts are present, and only falls back to `eta_seconds` otherwise (e.g. it matched `eta_seconds` during the download phase, which reports byte counts differently).

Two estimators disagreeing means one of them is always wrong on screen. Pre-PR-#2613 the payload ETA was the broken one (rate-extrapolation over a mispaced bar span); post-#2613 the payload's overall ETA converges quickly, so the frontend should probably display `eta_seconds` (the whole-job estimate) consistently, or at minimum the two should be reconciled — a phase-local "time left in this step" and a whole-job "time left" are different numbers and the UI should be explicit about which it shows.

Repro: import a demo dataset with a long step-2 decode (e.g. GTZAN warm-archive re-import), compare the row ETA against `eta_seconds` on `/api/events`.

Follow-up from #2556 / PR #2613.

samggreenberg:
## Summary

Fixes #2615 — the loading-row ETA (`eta_seconds`) was wildly wrong on warm and uncalibrated loads. Two root causes, both fixed here, plus the issue's UI-labeling ask.

First, the confirmation the investigation handoff asked for: **the dashboard row and the SSE payload cannot disagree — confirmed live.** Captured the browser row text and the raw `loading-tasks` frames at the same instants during a warm GTZAN re-import on the GRID dev: row "10 sec left?" ↔ `eta_seconds=12.4`, row "1 min left?" ↔ `eta_seconds=66.2`. The row is `formatEta(eta_seconds)`, frame-for-frame. The bug was always the payload value itself. (The original "row 2.5 min vs SSE ~3300 s" observation paired numbers from different instants of a pre-#2613 load — 3300 s is exactly the pre-pacer ballooned estimate from that session.)

### Root cause 1 — tracker: job-global rate extrapolation (`ProgressTracker._compute_overall`)

`eta = elapsed·(1−raw)/progressed` measured **since job start**. A warm re-import completes download+extract in under a second, banking ~30% of the bar instantly — the estimator reads that as sustained speed and projects "10 sec left?" at the start of a 2.5-minute embed, creeping upward the whole way (live GTZAN run: eta 12s → 66s → ~98s while true remaining went 157s → 21s, crossing from 9× under to 5× over).

**Fix:** rebase the rate window at every forward step boundary. The boundary update still samples the outgoing step's window (a fully-observed step is honest signal; the existing `_ETA_MIN_ELAPSED` gate discards the instantly-completed case), and the EMA carries across so the displayed value stays continuous. Existing ETA tests pass unchanged.

### Root cause 2 — pacer: seconds written into pseudo-time weights (`AdaptiveLoadPacer._observe_phase_rate`)

When no cost-model row matches (`load_cost_terms` static fallback — which is what every demo import without an explicit `media_type`/`embedder` gets), the terms are **unitless pseudo-times** (e.g. `load=0.10`, `embed=0.55`). The rate observation replaced the current phase's term with its projected duration **in seconds** — a 30 s model load vs `0.55` hands the observed phase ~100% of the remaining bar. Observed live (warm esc50_s, CPU): the bar leapt 0.115 → **0.977** the moment step 2 was observed, then the entire 100-second embed phase crawled 0.98 → 1.0 showing **eta < 1.5 s throughout**.

**Fix:** `load_cost_terms` now returns `(terms, calibrated)`. In pseudo mode an observation fixes the seconds-per-pseudo-unit **scale** — not-yet-observed phases are rescaled to keep their prior ratio to the observed phase. Calibrated (cost-model seconds) behavior is unchanged: only the observed phase's term is replaced.

### UI labeling

The `vt-job-progress` eta chip gets a hover title — "Estimated time until the whole job finishes, not just the current step" — so a whole-job number next to per-step counts stops reading as wrong.

## Validation (in-process on the GRID, warm esc50_s import, CPU static fallback)

| t | phase | before: overall / eta | after: overall / eta | true remaining |
|---|-------|----------------------|---------------------|----------------|
| 36s | embed 0/245 | 0.980 / 1.4s | 0.133 / 165s | ~103s |
| 75s | embed ~113/245 | 0.989 / 0.9s | 0.472 / 72s | ~56s |
| 122s | embed 225/245 | 0.998 / 0.3s | 0.807 / 27s | ~10s |

Bar paces through the dominant phase instead of pinning at 98%; ETA tracks the true remaining time and converges from ~1.6× to ~2.7× near the end of the phase instead of being off by 100×. Known residual: mid-model-load the ETA can briefly overshoot (~17 min shown at t=19s) until the phase completes ~10 s later — it is the honest "current pace continues" estimate with very little signal.

Tests: `tests_lib/core` + `tests_lib/datasets` (751 passed), new regression tests for both root causes and the pseudo-scale bridging; `job-progress.component.spec.ts` (9 passed) for the tooltip. Ruff clean.

## Deploy note

The dev server on rack8n06:11850 was restarted at 15:28 **from the main checkout before `git pull` picked up #2613** — it is still running pre-pacer code (that's what made today's live runs reproduce the pre-#2613 symptoms so faithfully). The checkout is now up to date; the next `vtsearch` launch picks up #2613, and this branch once merged.

### DIFF
--- a/frontend/src/app/components/job-progress/job-progress.component.html
+++ b/frontend/src/app/components/job-progress/job-progress.component.html
@@ -44,6 +44,6 @@
       [indeterminate]="bar().indeterminate"
       [smooth]="smooth()"
     />
-    <span class="jp__eta">{{ eta() }}</span>
+    <span class="jp__eta" [title]="eta() ? etaTooltip : ''">{{ eta() }}</span>
   </div>
 </div>
--- a/frontend/src/app/components/job-progress/job-progress.component.spec.ts
+++ b/frontend/src/app/components/job-progress/job-progress.component.spec.ts
@@ -33,6 +33,15 @@ describe('JobProgressComponent', () => {
     expect(text).toContain('5.5 min left');
   });
 
+  it('labels the eta chip as a whole-job estimate on hover', async () => {
+    const el = fixture.nativeElement as HTMLElement;
+    // No eta -> no tooltip (an empty chip should not invite hovering).
+    expect(el.querySelector('.jp__eta')?.getAttribute('title')).toBeFalsy();
+    fixture.componentRef.setInput('eta', '5.5 min left?');
+    await settleZoneless(fixture);
+    expect(el.querySelector('.jp__eta')?.getAttribute('title')).toContain('whole job');
+  });
+
   it('shows the info chip only when a description is provided', async () => {
     const el = fixture.nativeElement as HTMLElement;
     expect(el.querySelector('.jp__info')).toBeNull();
--- a/frontend/src/app/components/job-progress/job-progress.component.ts
+++ b/frontend/src/app/components/job-progress/job-progress.component.ts
@@ -38,6 +38,12 @@ export class JobProgressComponent {
   readonly detail = input('');
   /** Right-justified status, e.g. "5.5 min left" or "45%". */
   readonly eta = input('');
+  /**
+   * The chip shows the backend's whole-job estimate (`eta_seconds`), never a
+   * "time left in this step" number — say so on hover, since a long remaining
+   * phase can make it read as wrong next to the current step's counts (#2615).
+   */
+  readonly etaTooltip = 'Estimated time until the whole job finishes, not just the current step';
   /** Bar fill state; defaults to an indeterminate spinner. */
   readonly bar = input<ProgressBarState>({ value: 0, max: 1, indeterminate: true });
   /** Ease the fill for multi-phase jobs (see `vt-progress-bar`'s `smooth`). */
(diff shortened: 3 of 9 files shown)


samggreenberg:
Full non-GPU suite in the worktree: 6441 passed, 1 failed — the failure is `test_wavex_is_not_readable_by_stdlib_wave`, which fails identically on clean dev in this environment (pre-existing, filed as #2629, unrelated to this change).

PR #2795, opened 2026-08-01
Threshold-stability study (#2790): Stage-A replay + selectable threshold rules

samggreenberg:
Harness for the pre-registered threshold-stability study (`docs/plans/threshold-stability-experiment.md` on `dev`), issue #2790. Real Grid runs are owner-gated; the code + offline tests + CPU Grid runner are here.

## Headline: plan vs. this branch
The plan assumes the sod region-voting path "already uses the production conformal rule." **It doesn't** — `evaluation-framework` predates #2784, so *both* threshold paths reduce to min-cost **argmin**. Per the owner's steering ("match a simulation of Autopilot over the plan"), `argmin` = the current infidelity, `conformal` = fidelity-correct (what production Autopilot runs). Reframe confirmed by the owner.

## What's here (all tested; default byte-identical)
- **`vtscore/eval/threshold_rules.py`** — selectable, head-agnostic rules: verbatim `conformal_threshold` port (#2784), `stratified_fold_orderings`, `calibrated_threshold(rule=)` (`argmin`≡`xcal` byte-identical), `rank_transfer_cut` (S3), `median_smooth` (med3). — 16 tests.
- **`scripts/sod/replay_thresholds.py`** — Stage-A frozen-trace replay: cache-free tested core (vote reconstruction, per-step × rule × seed matrix, variance decomposition) + `CacheVectorLoader` (VALIDATE-ON-GRID). — 5 tests.
- **Stage B live-loop wiring** — `sweep.py` `--threshold-rule` / `--threshold-smooth`, threaded through `evaluate_realistic_curve` → `_train_pool_head` (whole path dispatches `calibrated_threshold`); med3 + diagnostic trace columns (`delta_threshold`, `spike`, `regret`). — 42 `test_region_curve.py` pass (byte-identical default).
- **CPU Grid runner** `scripts/experiments/threshold_stability/` — 6 arms × 5 classes × 10 seeds; Stage B sweep-per-arm + Stage A replay; CPU SLURM scripts; reuses the #2790 SigLIP2/whole cache.
- `scripts/sod/THRESHOLD_STABILITY_STATUS.md` — status, the reframe, and the remaining owner-gated work (validate `CacheVectorLoader` on one cell; region-voting-path conformal; run + report).

63 targeted tests pass; 172/172 `tests_lib/sorting`; ruff/codespell clean. Note: `sweep.py` carries pre-existing pyright PIL-monkeypatch warnings (lines ~179-192), not introduced here.

Refs #2790.

[...]

Issue #2790, opened 2026-07-31
MLP Threshold Calibration: COCO stop sign SigLIP 2 experiment


MatthewELucio:
<img width="1170" height="780" alt="Image" src="https://github.com/user-attachments/assets/052f9448-a10c-4926-8a0f-447c49a880a7" />

<img width="1170" height="780" alt="Image" src="https://github.com/user-attachments/assets/8822b929-81e3-4992-82db-acb9a7ac6957" />

I ran an experiment on COCO stop sign class with a SigLIP 2 embedder on the whole image, where different seeded runs are color-coded to check where the large variance in larger experiments is coming from: it seems to be related to the MLP threshold changing dramatically (oracle does not change much so the sorted ranking of the MLP output seems consistent). 

For example, in seed 0, at t=24, the cost jumps from 0.088 to 0.424 and F1 jumps down from 0.623 to 0.05 as the threshold drops from 0.49 to 0.35 in one training iteration, until stabilizing again. In seed 1, between t= 25 and 40, the cost and F1 have multiple divergences, interestingly showing opposite trends (when one improves the other gets worse).

To recreate the experiment in evaluation-framework branch run:

python scripts/sod/sweep.py --datasets coco --classes "stop sign" --cache-dir docs/experiments/sod-sweep/cache --embedders siglip2 --proposals whole --iterations 3 --max-labels 60 --out-dir docs/experiments/sod-sweep/coco-stopsign-siglip2-whole-3-iterations-oracle --viz --viz-band all --min-box-frac 0.03 --labeling-trace --show-oracle

samggreenberg:
Pre-registered experiment design in #2791 — `docs/plans/threshold-stability-experiment.md`. Plan only; harness knobs, the frozen-trace replay tool, and the Grid run are the next steps.

Likely root cause worth knowing now: the `whole`-proposal path of the sweep picks its threshold with the **old min-cost argmin averaged over 2 folds** (`vtscore/eval/xcal.py`), the exact rule production replaced with the conformal quantile + gap-midpoint cut in #2693/#2784 because the argmin jumps with a single changed held-out vote. The region-voting path already uses the production rule; the `whole` path (this issue's config) doesn't — so part of the observed instability may be a harness fidelity artifact rather than production behavior. The plan A/Bs that directly, plus fold count, temporal median smoothing, and a rank-space transfer arm for the fold→final score-scale mismatch.

---
_Generated by [Claude Code](https://claude.ai/code)_

[...]

PR #2796, opened 2026-08-01
Threshold-stability (#2790): Stage B results, report + spike attribution

samggreenberg:
Follow-on to #2795 (which merged the harness core). This is the **run + results**: the study was executed on the Grid (COCO × SigLIP 2 × whole, 6 arms × 5 classes × 10 seeds) and drilled into the individual spike-causing votes.

## Report
`docs/experiments/threshold-stability/REPORT.md` (force-added; `docs/experiments/*` is gitignored).

## Stage B verdict
- The cross-seed variance is **threshold placement, not ranking** (cost sd ≈ 4× the oracle floor).
- **Fold count (`k8`) is the dominant stability lever, not the rule.**
- **Conformal's win is regret (−37%)**, not jitter. No arm clears the strict adoption bar → S3 (fold→final scale) follow-up. Practical rec: **conformal-k8**.

## Spike attribution (278 up-spikes on argmin-k2)
- **87% caused by a Bad vote; 79% Bad + `hard`-selected** at the boundary (surfacing margin ≈ 0.001).
- **76% FNR-driven; 37% runaways** (test FNR up toward 1.0); **67% in the sparse-positive regime** (median n_good = 4).
- Only 26% narrow — 54% persist 5+ steps (compounding via boundary acquisition), correcting the "narrow blip" intuition.
- **Blockable:** the #2784 conformal within-model positive-coverage floor structurally blocks the runaway third; plus a sparse-positive deferral and an acquisition guard.

## Methodology note (from review)
The MLP is retrained every vote, so raw thresholds are **not comparable across steps**. All findings are stated on **test-set** cost/fnr/fpr and "far from *this* model's own oracle cut" — no cross-model threshold deltas. The per-vote-threshold-clamp idea was dropped as unsound for that reason.

## Also
- Fidelity fixes caught by the Grid pilot: replay now reuses the sweep's `MLPHead` (sigmoid scale); `class_slug` matches `features.slugify` (hyphens).
- Disk-safe: `--no-trace-images` writes `trace.json`/`trace.csv` without the PNG dump (the PNGs filled the 50 GB `/exp` and killed the first run); Stage B analysis derives everything from `results.jsonl`.
- `scripts/experiments/threshold_stability/` runner (cell=class) + `spike_analysis.py`.

Deferred (recorded in report + status doc): Stage A replay run, region-voting-path conformal, the S3 arm.

Refs #2790.

### DIFF
--- a/docs/experiments/threshold-stability/REPORT.md
+++ b/docs/experiments/threshold-stability/REPORT.md
@@ -1,0 +1,195 @@
+# Threshold-stability study (#2790) — report
+
+**BLUF.** On the COCO × SigLIP 2 × `whole` Autopilot labeling loop, the large
+cross-seed variance #2790 reported is **threshold-placement noise, not ranking
+noise** (across-seed cost sd ≈ 4× the oracle floor, every arm). The biggest lever
+on that noise is the **fold count** (`calibrate_count` 2→8 roughly halves the
+threshold jitter), not the calibration *rule*. Switching the sweep's historical
+min-cost `argmin` rule to production Autopilot's split-conformal rule barely changes
+threshold sd at matched fold count — but it **cuts calibration regret ~37%** (the
+cut is placed much better). No arm clears the pre-registered adoption bar; the
+best practical arm is **`conformal-k8`** (lowest spike rate and near-lowest
+threshold sd, with conformal's low regret). The residual points at the
+fold→final-model score-scale mismatch (S3), which needs a follow-up.
+
+## Setup
+
+- **Loop:** the realistic Autopilot labeling loop (`scripts/sod/sweep.py`,
+  `vtscore/eval/region_curve.py`), whole-image / box-pool path, MLP head.
+- **Config (the #2790 repro):** COCO, SigLIP 2 (`siglip2`), `whole`,
+  `--max-labels 60`, `--neg-multiple 100` (prevalence ≈ 1/101),
+  `--min-box-frac 0.03`, inclusion 0, `--safe-thresholds` on.
+- **Classes (5):** stop sign, traffic light, fire hydrant, parking meter, bus.
+- **Seeds:** 10. **Cells:** 6 arms × 5 classes = 30.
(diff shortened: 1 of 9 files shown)


samggreenberg:
## Argmin retired — backported #2784 conformal everywhere

Per review ("we don't even calculate argmin"), the eval is now all-conformal:

- **Backport:** `conformal_threshold` moved to `vtscore/training/thresholds.py` (its dev home); `threshold_from_fold_orderings` now applies **pooled conformal** instead of per-fold argmin. So the whole path, the region-voting/hac path, **and `vtscore/detectors/training.py`** (detector training on this branch) all calibrate with the shipped conformal rule.
- **Defaults** flipped to conformal (`region_curve`, `sweep.py`); `argmin` selectable only to reproduce old runs. Arms renamed `k2`/`k8`/`k2-med3`/`rank-transfer`.
- **Tests:** rewrote the fold-aggregation tests for the pooled-conformal contract; fixed `test_different_seeds_differ` (conformal makes *separable* data seed-independent → now uses overlapping data). The other cascade failures are **pre-existing** — `onnx` not installed (×10) + a known 2.3e-10 float32 flake — confirmed identical at the pre-backport commit. 1991 pass.
- **Report rewritten all-conformal.** Stage B verdict: raise `calibrate_count` 2→8 (`k8`) — the one lever that cuts cross-seed cost sd + spike rate; `med3`/rank-transfer don't help; residual → S3.
- **Spike attribution re-run on conformal traces (235 spikes):** 93% Bad vote, 89% Bad+boundary, **87% sparse-positive (median n_good = 3)**, runaways down to 22% (the gap-midpoint cap). The residual is a **too-few-positives** problem, not a rule problem — fix by deferring the trained cut / widening the coverage floor while positives are sparse.

Note: `./run-tests.sh` runs a pyright + full gate; the `onnx`-dependent detector-export tests need `onnx` installed to pass (pre-existing environment gap, not introduced here).

Issue #2841, opened 2026-08-04
HOW safe should safe-thresholds be?


samggreenberg:
We did an experiment to show that turning safe-thresholds on is better across the entire problem-space. But safe-thresholds has a hidden flexibility of HOW safe to make it. E.g. we could just use the GMM all the time every time. That's probably not better than ignoring the learned threshold entirely, right? Let's check.
Develop a variety of mix-in strategies to go from pure GMM (at the beginning) to pure learned thresh (eventually?). Then use the cached VG and COCO embeddings on the GRID to experimentally find the best plan.

samggreenberg:
Addressed in #2849 — full report at `docs/experiments/mixin-schedule/REPORT.md`.

**Short version:** the handoff was too fast, and production's cut turned out to be the worst-calibrated of the nine schedules measured (largest bias *and* largest spread against the oracle, on both voting modes). The two modes want different curves, as you suspected: region voting ships `slow` (hand off at 40 labels instead of 20, −0.0422 cost, p=1.4e-28), binary voting ships `cap50` (the old ramp but capped at half GMM forever, −0.0173, p=7.6e-43).

**On your premise:** "just use the GMM every time" is *not* obviously worse than ignoring the learned threshold — on region voting it was among the best schedules at the shipped operating point, and it sits closest to the oracle cut of anything measured. It isn't shipped because it's the least robust: weight false positives 4× and it swings from −0.023 to **+0.245** on binary voting. So the answer to "how safe" is *permanently a little safe*, not *always maximally safe*.

Two things I'd flag for you:
- **Roughly half the benefit is acquisition feedback.** Past 20 votes, where the blend has no authority left, the gaps are *bigger* than inside the ramp. The blend steers Autopilot's Hard pick, so a better cut buys better votes that keep paying off.
- **A real app/framework divergence turned up on the way in** and is fixed here: both app paths hard-coded "skip cross-calibration below 6 votes" while the harness never did, and the two paths disagreed on the placeholder they left behind — which a degenerate GMM makes user-visible (Find admits nothing, the vote path admits everything ≥0.5).

Known gap: COCO's cached embeddings lack the raw patch grid, so it could only serve the binary-voting arm — the region verdict rests on Visual Genome alone. Re-embedding COCO with patch grids is the top follow-up.

Issue #2799, opened 2026-08-01
Measure the safe-threshold GMM blend under region voting (calibration harness ran with safe_thresholds off)


samggreenberg:
**Recommended Claude model: Opus 4.8.** Experiment design and interpretation on the existing calibration harness — regression-prone metric plumbing plus judgment calls about what the cells mean; the harness itself already exists, so Fable-tier is not needed.

## Gap

The #2781 calibration study (`docs/experiments/calibration/REPORT.md`) ran every arm with `safe_thresholds False` (line 75), so all of its grouped-path bias findings concern the **conformal/cross-calibration** path only. The safe-threshold GMM blend (`calculate_safe_threshold`, `vtscore/training/thresholds.py:847`) — which has *full* authority below 6 votes and majority authority up to ~13 — has never been measured under region voting. That sub-20-vote window is also the cold-start cell the inclusion-calibration plan already flags as the measured weak spot (`docs/plans/inclusion-calibration-bias.md`, "Cold-start calibration").

## Work

Add a `safe_thresholds` arm to the existing harness (`scripts/experiments/calibration/`, driving `vtscore/eval/voting_iterations.py` — whose `_blend_safe_threshold` already scores region-aware) on the Visual Genome region-voting setup, and report FNR/FPR vs. vote count for:

1. **Status quo**: GMM blend as shipped (midpoint-of-means cut).
2. **Post-#2797**: blend fit on region max-pooled scores (the eval harness path already does this; the interesting comparison is production's image-level-score variant vs. the pooled one, to size the bias #2797 removes).
3. **Equal-density-crossing cut** (#2798), re-pooling the same fitted models where possible.

The 6-20-vote cells are the ones that matter (above 20 the blend is pure cross-cal and the existing study already covers it). Outcome decides whether #2798 ships, and whether the cold-start item in `docs/plans/inclusion-calibration-bias.md` should absorb a GMM-specific note.

Refs #2797, refs #2798.

samggreenberg:
Heads-up on the arm labels: #2798 (equal-density-crossing cut) landed in #2801, and #2797 (fit the blend on region max-pooled scores) landed in #2800. Both shipped ahead of this measurement, so when the harness arm is built:

- arm 1 "status quo (midpoint-of-means)" is now the **historical** cut, not what production runs;
- arm 2 "post-#2797 pooled fit" is on `dev`;
- arm 3 "equal-density crossing" **is** the current production rule.

[...]

PR #2849, opened 2026-08-04
How safe should safe-thresholds be? (#2841)

samggreenberg:
## What this is

Safe-thresholds has been on for everyone since #2799/#2835, but the schedule
*inside* the blend — how much GMM, for how long — had been one hard-coded line
since it was written:

```python
x-cal weight = clip((n_labels - 6) / 14, 0, 1)   # pure GMM ≤6, pure x-cal ≥20
```

Three choices were baked into it and none had been measured: the **endpoints**,
the **shape**, and the **statistic** it ramps on. A fourth — whether a weighted
average is the right combiner at all — couldn't be expressed in it. This makes
each one a named schedule, measures 18 of them, and ships the winners.

Full write-up: `docs/experiments/mixin-schedule/REPORT.md`.

## Verdict

| | ships | vs the old ramp (7–20 votes) | holds at `fpr x4`? |
|---|---|---|---|
| **Region voting** | `slow` — hand off at 40 labels, not 20 | −0.0422 cost, p=1.4e-28, 79% of cells | yes, −0.0366 |
| **Binary voting** | `cap50` — the old ramp, capped at half GMM forever | −0.0173 cost, p=7.6e-43, 68% of cells | yes, −0.0119 |

**Production's cut turned out to be the worst-calibrated of the nine measured** —
largest positive bias *and* largest spread against the oracle on both modes. The
shipped ramp hands over to a systematically conservative, unstable estimate too
early.

**The issue's premise was inverted.** "Just use the GMM always" is not obviously
bad — on region voting it was among the best at the shipped operating point and
is the closest to the oracle cut of anything measured. It is not shipped because
it is the *least robust*: weight false positives 4× and `pure_gmm` goes from
−0.0230 to **+0.2446** on binary voting. Since the Inclusion knob is defined as
`wf·FPR + wn·FNR`, that reweighting approximates a real user, not a hypothetical.

## Two things worth a reviewer's attention

**Acquisition feedback is about half the effect.** Past the ramp at 21+ votes,
where every schedule has converged and the blend has no authority at all, the
gaps are *larger* (region `pure_gmm` −0.1015, `vslow` −0.0968). Only the blend
having steered which items Autopilot asked the user to label can carry that. It
is invisible to any within-step analysis, which is why the cheap screen was
never allowed to decide.

**Why the winners win.** Decomposing each schedule's gap to its step's own
oracle cut into bias and spread: the cap family buys **spread** (a steadier cut,
which improves both error types under any weighting) while the slow schedules
buy **bias** (a lower cut, which only helps when the weighting likes it).
Averaging with a label-free GMM cut *is* variance reduction, so this was a
prediction the data then confirmed rather than a story fitted afterwards.

## App/framework fidelity — the deviation this found

Both app training paths hard-coded *"skip cross-calibration below 6 votes"*; the
harness never skipped. Safe only because the production weight happens to be 0
there — a coincidence, not a guarantee. The skip now derives from the schedule
(`xcal_is_discarded`), and the harness resolves its default from the same helper
the app uses, so the two agree by construction. Two real defects fell out:

- **The two app paths left different placeholders** (`NO_GOOD_THRESHOLD` vs
  `0.5`). Normally discarded — but a degenerate GMM makes the blend fall back to
  it, so the same collection would admit *nothing* through Find and *everything
  scoring ≥0.5* through the vote path. Both now admit nothing. Regression test
  included, and verified to fail when the divergence is reintroduced.
- **An off-by-one against its own rationale**: the guard skipped *below* 6 but
  paid for two 200-epoch fold fits at *exactly* 6, where the weight is already 0.

## Scale

12,096 A/B cells (nine arms × 16 seeds × 42 categories, 312,606 rows) plus a
1,008-cell screen, on the production linear head. Visual Genome + **COCO-2017-val**,
both from cached embeddings — nothing was re-embedded. COCO reaches the
calibration harness for the first time here (`build_coco_pickle.py` joins the
#2790 sweep cache to the staged boxes).

Fidelity check run before any number is reported: the `prod` counterfactual row
reproduced the live blended threshold and cost to **0.000e+00 over 26,142 paired
rows**.

## Limitations, stated in the report

- **COCO carries only binary voting** — its cache lacks the raw patch grid
  `max_patch` pools over, so the region verdict rests on VG alone. Weakest link.
- The `fpr x4` reweighting was **post-hoc** and is a *scoring* sensitivity, not a
  simulation of a different Inclusion setting.
- The pre-registered rule alone would have selected `vslow`; the recommendation
  differs *because* of the post-hoc robustness analysis, and says so.

`./run-tests.sh` equivalent on the GRID: **7634 passed, 51 skipped**.

Refs #2841


samggreenberg:
## Follow-up in flight: the horizon problem

Owner pushback, and it's right: `cap50` never handing over to the learned threshold cannot be literally correct, and **this study could not have seen why**.

The GMM midpoint is an **inconsistent** estimator of the cut — it reads no labels, so its error floors out at whatever its two-component symmetry assumption gets wrong, however much data arrives. The conformal cut is **consistent**. Asymptotically pure x-cal *must* win. So the question was never *whether* to hand over but **when**, and every number in this PR is measured over a **30-vote** horizon, where the answer is "not yet" almost by construction.

Running now:

- **Deep categories only** — a long horizon is bounded by *positives in the sim set*, not pool size; once autopilot exhausts them every further vote is a negative and the conformal positive-quantile stops improving. Floor of 50 sim positives keeps 14 VG / 12 COCO categories.
- **300 votes on binary voting**, 6 arms: `prod`, `cap50`, `pure_xcal`, plus three new cap-then-release arms bracketing the handoff (30→100, 50→200, 150→400).
- **200 votes on region voting**, 5 arms (region cells cost ~10× binary).
- Reported **per vote band** (7-20, 21-50, 51-100, 101-200, 201-300), because an average across a crossover is precisely the number that hides it.

**Read the binary recommendation in this PR as "best over the first 30 votes"**, which is where its evidence is. If a release arm beats both `cap50` and `pure_xcal`, the shipped binary schedule should become that arm and I'll push the change here before merge.

samggreenberg:
## Final: long-horizon results are in, and the region recommendation changed

**Shipped verdict:**

| | ships | vs the old ramp | at `fpr x4` |
|---|---|---|---|
| **Region voting** | **`slow_cap50`** (ramp to 40 labels, capped at half GMM) | −0.0577 (7-20), −0.094 by 51-100 | **improves** to −0.0745 |
| **Binary voting** | **`cap50`** (old ramp, capped at half GMM) | −0.0188 (7-20), −0.051 by 201-300 | **improves** to −0.0236 |

### The 30-vote horizon had picked the wrong region schedule

`slow` won the early window but **decays to +0.008 by 101-200 votes** — 40 labels is exactly where it becomes pure x-cal, so past that band it *is* the losing arm. `cap50` does the opposite (−0.026 → −0.097). `slow_cap50` is the synthesis: `slow`'s ramp with `cap50`'s cap. Best or tied in every vote band, wins 21-50 outright, and dominates `cap50` at **every** positive count.

### Does the learned threshold ever win? Not in a session anyone will have

Every cap-then-release arm gives its advantage back at the moment it releases, monotone in release point. Not because the GMM wins in the limit — it is an inconsistent estimator and cannot — but because **300 clicks buys a median of ~13 positives**, and the conformal cut converges in positives, not clicks. Banded on positives the convergence is real (`pure_xcal`'s deficit falls +0.096 → +0.014 on binary) and still ~100+ positives from parity.

### Region and binary fail differently

On binary `cap50`'s edge over `pure_xcal` **narrows** with positives (−0.111 → −0.048): the learned cut is converging, just far too slowly. On region it **widens** (−0.015 → −0.102) and `pure_xcal` flattens without ever crossing — the signature of a bias more labels cannot fix, plausibly the max-pool geometry (~196 patches) biasing the conformal cut structurally. Corroborating: `slow_cap50` is genuinely better calibrated on region (improves under `fpr x4`) but merely **more permissive** on binary (flips to +0.0414), which is why binary keeps `cap50`.

### Merge + verification

Rebased onto current `dev`, which meant integrating **#2842** (cut-rule decomposition) and **#2856** (hybrid calibration) — both landed in `voting_iterations.py` while this ran. Resolved by taking dev's rewritten machinery wholesale and re-applying this branch's changes on top. **7692 passed, 51 skipped** on the merged code.

One unrelated flake found and filed as #2858 (reproduces on clean `dev`, ~25% of runs).

### Also in this PR: ops machinery

A skill + `preflight.sh` + `LESSONS.md`, because these studies kept re-learning the same operational mistakes. `GRID-PLAYBOOK.md` already existed and was referenced from nowhere — the failure was discovery, not recording. The preflight **blocks** the checkable ones (results-dir collisions, free space on the real mount, stale worktree) and is wired into the launcher; it caught a real stale-worktree case on its first use.

PR #2863, opened 2026-08-06
Delete the safe_thresholds setting; the fused threshold is unconditional

samggreenberg:
Follow-up to #2861 (merged). That PR made the fold-anchored population cut the threshold whenever `safe_thresholds` was on; since it is better than the alternative at *every* label count, the toggle was only a way to opt into a worse cut. It's gone.

Every trained detector now fuses the haystack score distribution into its threshold, with nothing to configure.

## Removed end to end

- `CoreConfig.safe_thresholds`; `vtscore.state.get_safe_thresholds` / `set_safe_thresholds` and their persister registration.
- `GET|POST /api/safe-thresholds` and their marshmallow schemas; the `safe_thresholds` key on `/api/settings`; the `AppSettings` model field.
- The Settings → Sorting checkbox, its `sorting-api.service` bindings, and the generated API-client models (regenerated).
- The `safe_thresholds` parameter of `train_and_score`, `labelset_train_and_score`, `run_learned_sort`, `build_learned_sort_signature`, and `_train_and_score_xy`.
- `vtscore.training.thresholds.xcal_is_discarded` — it existed to skip the fold training where the schedule zeroed the cross-cal cut, and the estimator needs those fold models at every label count.

## What survives, and why

**The schedule blend is still the fallback.** `calculate_safe_threshold` and `vtscore/training/blend_schedules.py` still decide the threshold for label sets too small to form calibration folds — the regime where the schedule already degenerates to the pure GMM cut. Separately, `train_detector_from_origins` (re-deriving a saved detector with no dataset in hand) has no haystack to fuse and ships the cross-calibration cut alone.

**The eval harness keeps `safe_thresholds` as a research knob**, now defaulting to `True` so its baseline matches the app; `False` is the no-fusion control arm. That is a deviation that *is* the experiment, so it stays — and the paired `xcal_only` variant rows already measure the same contrast within-step. `eval_learned_sort` and `run_eval` lost the parameter entirely (and the `--safe-thresholds` CLI flag with it): they delegate straight to the production trainer, which has no such mode, so the knob would have been a lie. `_run_emit` in the calibration-harness tests now passes the control arm explicitly, since those tests pin the re-pooling plumbing and the conformal provenance ladder rather than the estimator.

## Breaking changes

For `vtscore` consumers: `CoreConfig` no longer takes `safe_thresholds`, and neither does `train_and_score`. `cross_calibration_threshold_cached` was already replaced by `calibration_folds_cached` in #2861. All recorded under `[Unreleased]` in `vtscore/CHANGELOG.md` alongside the threshold changes from the adoption itself — no version bump, since no release is being cut.

For the REST API: the two `/api/safe-thresholds` routes are gone and `/api/settings` no longer accepts or returns the key. A stored `safe_thresholds` value in an existing `settings.json` is simply ignored (`AppSettings` allows extra keys), so no migration is needed.

## Docs

`docs/ML.md` now states plainly that there is no setting and explains what the threshold computes; `docs/api/labeling.md`, `docs/api/settings.md`, `docs/ARCHITECTURE.md`, `docs/EVAL.md`, `docs/DEPLOYMENT.md`, `docs/vtscore-api.md`, and the `vtscore/docs/` library docs are updated. `docs/plans/safe-threshold-mixin-schedule.md` (#2841) gets a status note: its search still has a subject, but only the small-label fallback window, and the stakes shrank accordingly.

No screenshot reshoots queued — the only settings shot in the manifest frames the **Appearance** pane, and the side tab bar is unchanged.

`./run-tests.sh` green (7805 passed), `./run-tests.sh vtscore-clean` green (3699 passed).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Pb5aNunLUEpNkKrL8jt55P)_


samggreenberg:
Measured evidence against making this unconditional, from the anchor-mass sweep in #2864 (SLURM 468874, 384 cells, 6 environments, 0 failures).

That run scores the shipped fused threshold against the `slow_cap50`/`cap50` blend it replaced, as counterfactual rows on the same trajectory. Deep regime (51–300 votes), cell-paired, pooled within voting mode — negative = beats the blend:

| Arm | region voting | binary voting |
|---|---:|---:|
| `fold_anchored κ=0.3 mid` | −0.0256 | −0.0004 (n.s.) |
| **`fold_anchored κ=1 rate` (what #2861 shipped)** | −0.0234 | **+0.0063** |
| `xcal_only` | +0.0520 | +0.0103 |

On region voting fusion is a clear, highly significant win. On **binary voting the shipped setting is worse than `cap50`**, and on the smallest set measured (caltech101, 838 images, 3 positives in the deep regime) every fusion arm loses to `cap50` by +0.003…+0.025. Even the best fusion setting only reaches a dead heat there.

The pattern is not dataset size as such but **positive-anchor count** — deep-regime median positives vs deep Δregret at each environment's best: 24 → −0.093, 8 → −0.019, 7 → −0.068, 3 → −0.002. The mixture needs a Good component it can locate; `cap50` buys spread instead, which is what a threshold estimated from three positives is short of (the same mechanism #2841 found).

So deleting the setting would make a measured regression unconditional for binary-voting detectors. Two things would unblock it: fixing the constant (#2864 recommends `κ=0.3, mid`, which wins in 6/6 environments and is the only arm that doesn't regress binary voting), and a positive-count gate — "use fusion once the fold anchors contain ≥ k positives, else the blend". k isn't estimated yet.

Caveat: counterfactual schedule rows re-cut *this* trajectory, so they bound the threshold-rule difference, not the whole-system difference (#2841's screen caveat). Full detail in `docs/experiments/population-anchored-calibration/REPORT.md` on #2864.

Issue #2731, opened 2026-07-29
Production threshold calibration compares a max-over-1 Good bag against a max-over-13 Bad bag


samggreenberg:
**Difficulty:** hard — small diff, but it moves the decision threshold every patch-dataset detector uses.

**Recommended Claude model: Opus 4.8.** This is the shipped threshold path; a wrong-but-plausible change silently shifts every detector's operating point, and the failure mode is invisible in ranking metrics.

## The defect

On a patch dataset, `_build_vote_xy` (`vtscore/detectors/training.py:397-406`) builds asymmetric bags:

- a **Good** vote contributes **one** row, group `("g", cid)`
- a **Bad** vote contributes `bad_negative_vecs(...)` = the CLS node plus every HAC leaf, ~**13** rows, group `("b", cid)`

`_compute_fold_orderings_grouped` (`vtscore/training/thresholds.py`) then collapses each calibration group with `max`. So the threshold is fit by comparing positives scored as **max over 1 row** against negatives scored as **max over ~13** — while `_score_all_media` scores *every* image, positive and negative alike, as a max over all ~24 region nodes.

`max` is an upward-biased order statistic: 13 draws from the top of a distribution beat 1 draw even with no signal at all. Calibration therefore systematically understates what a positive will score at inference relative to a negative, and the min-cost cut lands high. The bias is toward **rejecting true matches** (FNR), which is the expensive direction for a search tool.

The bag-aware calibration path was added precisely so a Bad bag "scores as one image, as at inference" — this issue is that the Good bag never got the same treatment.

## How it was found

While investigating why the Max-Patch experiment's Caltech-101 arm collapsed (#2730). The experiment tier had a far more extreme version of this — max-over-1 vs max-over-196 — where it was catastrophic rather than merely biased. Instrumenting the calibrator with one model held fixed showed the Bad side matching production exactly (shift +0.0000) and the Good side not matching at all.

Production is milder for two reasons: the ratio is 1:13 rather than 1:196, and the Good vote's vector is always a real scoring row (`pool_box_from_media` snaps to a tree node, and the boxless fallback is the CLS vector which *is* `patch_regions[0]`). So this is a threshold-placement bias, not the geometry mismatch that broke the experiment. It is still worth fixing.

## Suggested approach

[...]

PR #2894, opened 2026-08-07
Adopt MaxPatch as the region-vote strategy; drop the HAC region tree from ingest

### LINKED ISSUE #2886: Adopt MaxPatch as the region-vote strategy and drop the HAC region tree from ingest
**Difficulty: hard.** A cross-cutting change to the vote/score geometry that touches training, the scoring matrix, ingest, the pickle payload, the cross-dataset labelset path, and the best-region UI overlay. Wrong-but-plausible is expensive here: the last geometry slip in this area (a trained vector that was never a scored row) produced *perfect ranking with catastrophic FNR* and took a whole re-run to catch.

**Recommended Claude model: Opus 4.8.** The train/score geometry invariant and the calibration coupling are exactly the "subtle, regression-prone, costly if plausibly wrong" regime. The dead-code sweep at the end (step 8) is Sonnet 5 work if it gets split out.

## Why

The Max-Patch study's verdict — [`docs/experiments/max-patch/REPORT.md`](../blob/dev/docs/experiments/max-patch/REPORT.md) — is **"ship tree-free MaxPatch as the DINOv3 region-vote strategy, and drop the HAC tree from ingest."** That verdict currently lives only as prose in the report's "Plans for moving forward". No issue tracks it, and production still runs MaxHAC unchanged. This issue is that missing owner.

Over 23 scale-band Visual Genome categories × 3 seeds, at t = 150:

| arm | ErrorCost | AP | FPR | FNR |
|---|---|---|---|---|
| **MaxPatch** | **0.40** | 0.486 | 0.089 | 0.312 |
| MaxPatchHAC | 0.44 | **0.492** | 0.104 | — |
| MaxPatchPcaHAC | 0.42 | — | — | — |
| MaxHAC (production today) | 0.46 | — | — | — |
| DINOv3 CLS whole-image | 0.61 | — | — | — |

- MaxPatch beats production MaxHAC by paired Δ = −0.064 (Holm p = 0.002), and is best-or-tied-best in **every** scale band, on both halves of the error.
- The scale hypothesis holds: MaxPatch's edge is largest on small objects and shrinks as they grow (Spearman ρ = 0.50 between voted-box area and the MaxPatch−MaxHAC gap, p = 0.016). Below leaf scale a raw patch is a near-pure object sample while the tree's smallest k-means-pooled leaf already blends object with context.
- The tree does not earn its keep in either form: the *new* raw-patch-leaf tree (MaxPatchHAC) only numerically edges MaxHAC (Δ = −0.027, p = 0.064 n.s.) and loses to plain MaxPatch (Δ = +0.037 n.s.). PCA merge-ordering is a statistical no-op.
- Follow-up confirmation from the calibration study ([`docs/plans/calibration-experiment.md`](../blob/dev/docs/plans/calibration-experiment.md)): the raw-patch tree is provably *calibration*-bottlenecked — it ties or beats at the oracle threshold — and no pre-registered remedy (`topk`, sign-corrected `pnorm`) recovers it, "so `max_patch` stays the region-vote strategy."

## What production does today (the thing being replaced)

- **Ingest** — `vtscore/datasets/stages/embedding.py::_attach_patch_regions_to_media` runs `build_region_tree(patch_out, k=12, alpha=0.5)` → 24 nodes per image (CLS + 12 saliency-weighted k-means leaves + 11 HAC internals), stored fp16 as `media["patch_regions"]`. `media["patch_grid"]` is stored alongside it.
- **Good region vote → 1 row** — `training.pool_box_from_media` → `patch_embed.snap_box_to_region`: highest box-IoU node across the whole 24-node tree.
- **Bad vote → 13 rows** — `training.bad_negative_vecs` floods the childless nodes (CLS + the 12 leaves). Internals are deliberately excluded; see #2731.
- **Inference** — `embedding/matrix.py::_build_region_arrays` + `_score_all_media` max-pool the MLP over all 24 nodes.

## Target behaviour (MaxPatch)

Reference implementation already exists and is tested at experiment tier: `vtscore/eval/patch_styles.py::MaxPatchStyle`.

- **Good region vote** → the single raw patch nearest the box (`patch_embed.nearest_patch_to_box`, already in `vtscore/`).
- **Bad vote** → the full-image/CLS vector **plus every raw patch** (1 + H·W rows; 197 on DINOv3 14×14).
- **Inference** → max-pool over the full-image row stacked above all raw patches.

**The CLS row is load-bearing in all three, not decoration.** `patch_styles.py`'s module docstring states the invariant this change must not break:

> Every vector a style can train a vote on must also be a row that style scores over.

MaxPatch originally scored raw patches only, so a *boxless* Good vote — which falls back to the CLS vector — trained on a vector nothing ever scored; the classifier separated "full-image-like" from "raw-patch-like" and the calibrated threshold landed in a gap the score distribution never reaches. Any implementation that drops the CLS row from **either** `bad_vecs` or `score_rows` reintroduces that failure.

## Scope

- [ ] `vtscore/detectors/training.py` — `pool_box_from_media` → nearest-patch; `bad_negative_vecs` → CLS + all raw patches; `inference_score_rows` → CLS + flattened grid. Keep the bag tagging (`("g", cid)` / `("b", cid)`) so a Bad vote's ~197 rows still count as one image.
- [ ] `vtscore/embedding/matrix.py::_build_region_arrays` — rows come from `patch_grid` + CLS instead of `patch_regions`. `region_index_per_row` changes meaning (row 0 = whole image, rows 1.. = grid cells in row-major order); the winning index still has to resolve to a box for the UI overlay.
- [ ] `vtscore/datasets/stages/embedding.py::_attach_patch_regions_to_media` (and its `loader_folder._attach_patch_regions` mirror) — stop calling `build_region_tree`; keep storing `patch_grid`. **Note the payload gets smaller, not bigger** — the grid is already pickled, so MaxPatch needs strictly less stored data than MaxHAC.
- [ ] `vtscore/detectors/labelset_training.py` — the cross-dataset resolve path builds a fresh tree (`build_region_tree(output)`) then snaps; becomes nearest-patch on the re-derived grid.
- [ ] `vtscore/training/region_similarity.py` — exemplar/text seeding max-cosines over region nodes and returns a `best_region_box`; becomes max-over-patches, and the returned box is a patch cell.
- [ ] `vtscore/training/structural_similarity.py` — gates the patch path on `patch_regions`; needs a gate that survives the tree's removal.
- [ ] Frontend — `image-viewer.component.ts`'s `highlightBox` best-region outline becomes a single patch cell (or the whole image). Decide whether a one-cell rectangle is still a useful overlay or should be suppressed; `highlightBoxStyle` already suppresses the ≥0.99 whole-image case.
- [ ] Dead-code sweep — `build_region_tree`, `build_hac_tree`, `propose_leaves`, `snap_box_to_region`, `box_to_vote_vector`, `RegionVector.cell_mask` / `.weight`, plus their tests. The #2731 internal-flood gap tests (`tests_lib/detectors/test_hac_internal_flood_gap.py`, `scripts/probe_hac_internal_flood.py`) become moot — delete rather than leave asserting a geometry that no longer exists.
- [ ] Perf — scoring goes from ~24 rows/image to ~197 (≈8×); the flattened matrix should stay fp16 with chunked upcast, as `patch_styles._forward_sigmoid_chunked` already does. Ingest *gains* (the O(k³) HAC build goes away). Measure both on the largest collections before calling it done.

Existing pickles carry both `patch_regions` and `patch_grid`, so old datasets keep working by simply ignoring the tree. Per `CLAUDE.md`, no compatibility shim is wanted — just stop reading it.

## Deliberately out of scope

- **The 4-DOF question.** `nearest_patch_to_box` collapses to "the patch nearest the box center" — the drawn box's width and height are discarded in essentially every case (the in-box restriction only binds for a box thinner than one cell). That is the shipped MaxPatch design and what the study measured; do not "improve" it here. The natural alternative (`mean of patches inside the box`) is an unrun optional arm in [`docs/plans/max-patch-experiment.md`](../blob/dev/docs/plans/max-patch-experiment.md), and it violates the train/score invariant above by construction, since a per-vote amalgam can never be a per-image scored row.
- **Max-pool-aware calibration for the raw-patch tree.** MaxPatchHAC ranks best of any arm (AP 0.492) and recovers real large-object recall; it loses only at the operating point. That lever is already tracked as an open follow-up in `docs/plans/calibration-experiment.md` and should not block this adoption.

Refs `docs/plans/max-patch-experiment.md`, `docs/experiments/max-patch/REPORT.md`, #2730, #2731.

samggreenberg:
Acts on the Max-Patch study's verdict — ship tree-free MaxPatch, drop the HAC tree from ingest.

Closes #2886

## The geometry

One definition now owns it: **`vtscore.embedding.matrix.media_score_rows`** — row 0 is the image-level (CLS) vector, rows `1..H*W` are the raw patches row-major. `patch_row_box` is the inverse map back to a box.

| | before | after |
|---|---|---|
| ingest | `build_region_tree(k=12, α=0.5)` → 24 nodes + grid | grid only |
| Good region vote | snap box to best-IoU tree node | `nearest_patch_to_box` |
| Bad vote | CLS + 12 childless leaves (13 rows) | `media_score_rows` (197 rows) |
| inference | max-pool over 24 nodes | max-pool over the same 197 rows |
| best-region overlay | pooled region box | one grid cell, or the whole image |

**The CLS row is load-bearing in all three.** The Good-vote path, the Bad flood, and the scoring matrix all call `media_score_rows`, so the train/score invariant holds by construction rather than by three implementations agreeing — a boxless Good vote trains on row 0, which the scorer evaluates.

Dropping the tree also closes the #2731 flood/score gap for free: internal HAC nodes were scored but never flooded (renormalised merge vectors aren't dominated by their own leaves). MaxPatch has no internals, so the exception is gone rather than inherited. `test_hac_internal_flood_gap.py` and `probe_hac_internal_flood.py` are deleted along with the geometry they asserted.

## Measured cost (DINOv3 14×14, D=768, CPU)

| | before | after |
|---|---|---|
| ingest, per image | 2.52 ms | 0.46 ms |
| scored rows per image | 24 | 197 |
| score matrix, per 10k images | ~740 MB f32 | ~3.0 GB f16 |
| forward pass, per image | ~13 µs | ~110 µs |

Ingest gets ~21 s cheaper per 10k images and the **payload shrinks** — the grid was already pickled alongside the tree, so the tree's ~150 MB per 10k images is pure saving. Scoring is where the cost went: a retrain runs three scoring passes (final model + one per calibration fold), so a 10k-image collection pays roughly 3 s per vote instead of 0.4 s. Bounded by keeping the flattened matrix **float16** with chunked upcast (`_forward_sigmoid_chunked`, `chunked_row_scores`), and by `_build_region_arrays` preallocating and filling in place instead of concatenating per-media blocks (which would hold 2× the matrix at peak). The matrix is still cached per media-id set, never rebuilt per vote.

## Changed

- `detectors/training.py` — `pool_box_from_media` → nearest patch; `bad_negative_vecs` / `inference_score_rows` → `media_score_rows`; `_format_results` → `media_row_box`; new chunked forward. Bag tagging unchanged, so a Bad vote's 197 rows still count as one image.
- `embedding/matrix.py` — `media_score_rows`, `media_row_box`, `chunked_row_scores`, `ROW_CHUNK`; `_build_region_arrays` rebuilt around the grid.
- `datasets/stages/embedding.py` — `_attach_patch_grid_to_media`; the back-fill gate moves from `patch_regions` to `patch_grid`.
- `detectors/labelset_training.py` — one `_resolve_score_rows` feeds both the negative and scoring caches (identical under MaxPatch), replacing the tree rebuild + `_leaves_from_regions` / `_all_region_vecs` pair.
- `training/region_similarity.py` — max-over-rows with a chunked matvec.
- `media/patch_embed.py` — 777 lines down to the wire types, the embedder adapters, `nearest_patch_to_box`, and `patch_row_box`.

## Frontend

No behaviour change needed. `highlightBoxStyle` already suppresses the ≥0.99 whole-image case, so a winning image-level row draws nothing; a winning patch draws one cell (~7% of the side on a 14×14 grid), which is genuinely informative — it points at the exact patch that scored highest. Comments updated to describe the new semantics.

## Deliberate deviations from the issue's checklist

- **`RegionVector` / `snap_box_to_region` / `_fit_pca_projector` were moved to `vtscore/eval/patch_styles.py`, not deleted.** The `max_patch_hac` / `max_patch_pca_hac` arms depend on them and are a *tracked open follow-up* in `docs/plans/calibration-experiment.md` ("max-pool-aware calibration for the raw-patch tree"). Production is fully tree-free; the tree survives only as experiment-tier code that owns its own node type.
- **The `max_hac` arm is deleted.** It delegated to the production code this PR removes, so keeping it runnable would have meant re-implementing exactly what the study told us to delete — and it has no tracked follow-up. Dropped from the runner grid with a note; its published numbers stand in the report and `analyze.py` still labels `max_hac` rows in archived CSVs.
- `scripts/run_hac_tree_sweep.py` and `scripts/inspect_hac_duplicates.py` are deleted too — both are sweeps over the tree builders that no longer exist.

`loader_folder._attach_patch_regions` (named in the issue) no longer exists; `stages/embedding.py` is the only ingest site.

## Backwards compatibility

Breaking, as the issue intends. `patch_regions` was never persisted — the pickle serializer writes an explicit field allowlist and the side-channel is re-derived at load — so old datasets load fine and simply gain no tree. No shim.

## Tests

`./run-tests.sh`: **7792 passed, 1 skipped**. Includes the full ruff / codespell / deptry / pip-audit / pyright / OpenAPI-snapshot gate and the Angular build.

One pre-existing failure worth naming, not fixed here: `tests_lib/detectors/test_acquisition_threshold.py::test_the_offset_is_relative_to_inclusion_not_absolute` fails when its file is run in isolation, on unmodified `dev` as well as on this branch — the anchored cut saturates at inclusion −6 so the strict `>` collapses. It depends on global RNG ordering and passes under the full suite. Unrelated to patch geometry and I'd be guessing at the intended contract, so I left it.

https://claude.ai/code/session_019R7AxLQjHi952HSgzAv71B

---
_Generated by [Claude Code](https://claude.ai/code/session_019R7AxLQjHi952HSgzAv71B)_


samggreenberg:
**Re: does the eval default (`style=None`) stay in line with the app default?**

Two of its three legs do, automatically, because they delegate to production rather than re-implement. The third was already out of line and this PR widens it.

| leg of the `style=None` path | implementation | tracks production? |
|---|---|---|
| Good vote | `_good_training_vec` → `pool_box_from_media` | ✅ now the nearest raw patch, for free |
| test / sim scoring | `_score_sim_set_with_model` / `_step_metrics` → `score_media_with_model` | ✅ now the ~197-row max-pool, for free |
| Bad vote | hardcoded `media_embedding(clips_dict[vid])` | ❌ one image-level row, as before |

Two things were load-bearing for the scoring leg and are in the diff:

- `region_aware` in `simulate_voting_iterations` used to gate on `patch_regions`. Left alone, patch datasets would have silently fallen through to whole-image scoring in the default arm once ingest stopped writing trees — a silent measurement regression, not a crash. It now gates on `patch_grid`.
- The same flag feeds `production_schedule_for(region_voting=...)`, so the blend schedule would have flipped to the `binary` arm on patch datasets too.

The Bad-vote divergence is pre-existing and deliberate (`docs/plans/max-patch-experiment.md`, Known limitations — the default path is frozen for reproducibility of pre-style studies), so I did **not** change it. But its magnitude moved a lot: the default arm now trains a Bad vote on 1 row while scoring over ~197, where before it was 1 against ~24. The train/score *invariant* still holds — the image-level vector is row 0 of the stack — but ~196 patch rows per rejected image are never trained down while inference max-pools them, so the arm will under-suppress on patch datasets more than it used to.

Pushed a doc update recording exactly that, with a pointer to prefer `style="max_patch"` for anything meant to reflect the live tool. Suite still green (7792 passed).

---

samggreenberg:
**Fixed — the eval default is now the app default on patch datasets.**

My earlier "left untouched for reproducibility of earlier studies" defence doesn't hold, and I should have caught that before documenting the divergence instead of fixing it. Two of the default path's three legs *delegate* to production (`pool_box_from_media`, `score_media_with_model`), so its Good vote and its scoring had already moved to MaxPatch geometry along with the app. Freezing only the Bad vote didn't preserve the historical harness — it produced an incoherent hybrid that matched neither the old production nor the new one:

| leg | old prod | style=None (as I'd left it) | new prod |
|---|---|---|---|
| Good vote | snap to node | **nearest raw patch** | nearest raw patch |
| scoring | 24 nodes | **197 rows** | 197 rows |
| Bad vote | 13 rows | **1 row** | 197 rows |

So it wasn't reproducing anything — it was training ~196 rows per rejected image down never while inference max-pooled all of them, i.e. systematically under-suppressing. Exactly the "evals can't be trusted" case.

**The fix** (`simulate_voting_iterations`): once `region_aware` is known, `style=None` on the MLP trainer resolves to `"max_patch"` — which *is* the production geometry, since `MaxPatchStyle`'s three methods delegate to `pool_box_from_media` / `bad_negative_vecs` / `media_score_rows`. That routes the default arm through `_style_train_and_calibrate`, so it also picks up the bag-aware machinery it was missing: per-bag loss weights, grouped calibration folds, and inference-geometry `score_rows`. No duplicated logic.

Untouched: single-vector datasets (no grid → no style → the historical `_mlp_train_and_calibrate` path byte-for-byte) and the non-MLP trainers (no torch head for a style to drive).

The **resolved** name is what lands in the `style` column, so a result row always says which geometry produced it rather than hiding the switch behind `""`.

Tests — replaced the old `test_default_run_records_empty_style` with four that pin the contract from both ends:
- patch dataset defaults to `max_patch` and records it;
- single-vector dataset still records `""`;
- **a style-less run is trajectory-identical to an explicit `max_patch` run** (the real contract, not just the label);
- a Bad vote's flood is `GRID*GRID + 1` vectors and byte-equal to `bad_negative_vecs` — so a refactor can't restore the 1-row Bad vote while still recording the name.

Also corrected `_mlp_train_and_calibrate`'s docstring, which claimed its whole-image Bad votes were "matching the live detector" — false since region flooding shipped, and the kind of comment that makes the next person reinstate the bug.

Suite green: **7795 passed**.

---

Issue #2897, opened 2026-08-07
Better Calibration => More Calibration?


samggreenberg:
We've made lots of improvements to calibration, in both binary- and region-voting. Cross-partiallyLabeled-GMM is great. But we haven't investigated the NUMBER of crosses in a while.

1) Write an experiment to evaluate the costs and benefits of increasing the number of folds. (I presume runtime will go up and oracle-regret will go down? By how much?) Do both binary and region datasets.

2) Run the experiment on the GRID.

samggreenberg:
Part 1 (write the experiment) is in #2902. Part 2 (run it on the GRID) is still open, and is tracked as open work in `docs/plans/calibration-fold-count-experiment.md`.

One design note worth surfacing here, since it changes what the run costs: the calibration folds are **not** a K-fold partition — each is an independent stratified split at a fixed `calibration_fraction`, drawn sequentially off one `RandomState(42)`, with `train_model` seeded per call. So the folds are *nested*, and the K folds a live `calibrate_count=K` run trains are byte-for-byte the first K of a Kmax run's. That means one instrumented run measures every fold count's regret **and** every fold count's wall clock, paired within the step, instead of needing one full run per K. It also means K is a variance knob rather than a data-efficiency one, so the pre-registered prediction is a saturating benefit against a cost that is linear in K.

The screen still can't see acquisition feedback (K steers the rank position Autopilot's Hard pick samples around), so there's a second stage of live A/B runs and the analyzer reports whether the screen agreed with them.

---
_Generated by [Claude Code](https://claude.ai/code)_

samggreenberg:
Part 2 (the GRID run) is now launched — see #2906.

Two things worth flagging before the results land:

- The **only genuine region-voting arm could not load at all** under current `dev`. #2886 removed `RegionVector` from `vtscore.media.patch_embed`, but the cached `visual_genome_m__dinov3_patch` pickle stores a per-image HAC tree of those nodes and `load_medias` is a bare `pickle.load`. Every region cell would have died before returning a media. Fixed in #2906.
- **`visual_genome_m × siglip` does not region-vote**: `patch_grid` on 0 of 4193 media, so `region_voting=True` silently falls back to whole-image (same trap as #2877). Both modes are still covered — `dinov3_patch` is genuine region voting at 4193/4193, `caltech101_m` genuine binary — but that arm is a third *binary* environment, not a region one.

[...]

Issue #2877, opened 2026-08-07
Generalisation check for the acquisition cut: re-run the k sweep on visual_genome_m × siglip region voting


samggreenberg:
**Difficulty:** medium — the harness knob, the launcher and the analyzer all exist; this is a second environment through machinery that has already run once.

**Recommended Claude model: Sonnet 5.** The mechanical work (repoint the grid, launch, drain, analyse) is bounded and well-specified. Step up to Opus 4.8 only if the result comes back *split* — a positive COCO answer and a negative VG one is a shipping decision, not an analysis, and that is where a wrong-but-plausible reading is costly.

## What

PR #2876 measured the acquisition/reporting threshold decoupling on `coco_val × siglip2 × whole_image` (binary voting), found the interior optimum at `acq_inclusion_offset = -3`, and that offset now ships as the default in both the app and the eval harness.

The plan that scoped the run pre-registered a **second environment, conditional on the first being positive**: `visual_genome_m × siglip`, region voting. It is positive, so the check is owed.

## Why it is not a formality

#2861 showed that this family of answers does not always transfer between voting modes. Everything behind the shipped default is COCO binary voting in a single embedder. Region voting differs in exactly the way that could matter here: a media's score is a max over ~24 region-node scores, so the Bad mode is an extreme-value statistic and the score distribution the acquisition cut is realized on has a different shape.

There is also a known interaction. `docs/ML.md` records that the fused threshold's gain tracks how many *positive* anchors the regime supplies, and that it wins clearly on region voting while being a dead heat on COCO binary voting. The acquisition cut rides on the same fitted mixture via `mid_tilt`, so region voting is plausibly where the offset does *more*, not less — but that is a hypothesis, not a result.

## Design

Reuse the #2876 arms verbatim so the two environments are directly comparable:

| arm | `CALIB_ACQ_INCLUSION_OFFSET` | role |
|---|---|---|
| `prod` | `0` | control — the pre-#2876 coupled behaviour |
| `acq_m1` … `acq_m4` | `-1` … `-4` | the frontier |
| `acq_p2` | `2` | **falsification arm** — must make positives worse |
| `rank_pin` | `0` (+ `CALIB_ACQ_RANK_PERCENTILE=0.959`) | the pinned-quantile parameterisation |

Keep `acq_p2`. It is the only thing that separates "the lever works" from "any perturbation of the sampling position changes the numbers", and on COCO it was what made the rest interpretable.

[...]

PR #2906, opened 2026-08-07
Fold-count study #2897: unblock the region arm, and chain the GRID run end to end

samggreenberg:
Runs #2897 part 2 (the GRID run). Part 1 landed in #2902.

Refs #2897

## The bug this had to fix first

The study's **only genuine region-voting arm** could not load. #2886 took
production tree-free and removed `RegionVector` from
`vtscore.media.patch_embed`, but the cached Max-Patch
`visual_genome_m__dinov3_patch` pickle stores a per-image HAC tree of exactly
those nodes, and `_cells_io.load_medias` is a bare `pickle.load`. So every
region cell would have died with `AttributeError` before returning a single
media — taking the region half of #2897 with it, on a run nobody was watching.

Fixed with a narrow compat shim, after confirming nothing under `vtscore/`
reads `media["patch_regions"]` any more. The placeholder is deliberately *not*
aliased to the surviving `vtscore.eval.patch_styles.RegionVector`: that is a
different, narrower dataclass (the old one also carried a `cell_mask`), and
equating them would be a lie a later reader could act on.

This affects **any** study reusing pre-#2886 patch pickles, not just this one.

## A premise that did not hold

`visual_genome_m × siglip` **does not region-vote**: `patch_grid` is present on
0 of 4193 media, so `region_voting=True` silently falls back to the whole-image
vector — the same trap as #2877. Both voting modes are still covered
(`dinov3_patch` is genuine region voting at 4193/4193, `caltech101_m` is genuine
binary), but that arm is a third *binary* environment and the plan's prose is
wrong about it. `prepare` prints the fact plainly (`patch grids on 0` vs
`on 4193`), so the plan is corrected to point at prepare's own output rather
than the dataset→`region_voting` table.

## Chaining, so the run finishes unattended

- `chain_folds_2897_ab.sh` reads the screen's `summary.json` and launches the
  live A/B **only when** some voting mode's `h3_recommended_k != 2`. When every
  mode reports `h3_kept_production`, it launches nothing — that null is the
  pre-registered result, and two more full simulations would only re-confirm it.
- `launch_folds_2897_ab.sh` gains `CALIB_AB_BASE`, so arms can land off the
  50 G `/exp` quota. An arm that runs out of disk loses its *late* steps, which
  is precisely where the saturation question lives.
- `status_folds_2897.sh` answers "is it done" from files rather than `squeue` —
  a drained queue and a finished study look identical to the scheduler. It
  counts cells against the expected total and names zero-byte cells that a
  resume would silently skip.

## Run

208 cells (92 VG×siglip, 92 VG×dinov3_patch, 24 Caltech), K ∈ {1,2,3,4,6,8,12,16},
4 seeds × 150 steps. Sized from real cells before submitting: VG×siglip 2:57 at
964 MB RSS and 1.25 MB output, so Kmax=16 fits the window with large margin.
Results on `/home` (no quota) rather than `/exp` (92% full, and 13 G of it is the
shared `.venv` that can never be reclaimed).

**The run is in flight as of this PR**; the report and the verdict follow.

## Testing

- `selftest_analyze_folds_2897.py` — OK
- Shim verified in the real code path: `load_medias` returns 4193 media, all
  with `patch_grid`, and the array's region cells are running on it.
- Shell scripts syntax-checked; `ruff check` / `ruff format` clean.

Note for a reviewer: `scripts/check-eval-app-sync.py` is **already red on `dev`**
(three autopilot progress mirrors, all with declared input-plumbing divergences).
Not touched here — re-pinning without reconciling is what that gate exists to
prevent — but it means `./run-tests.sh` fails on `dev` until someone looks.


samggreenberg:
## Run complete — 624 cells, zero failures

Screen 208/208, A/B arms 208/208 each, no zero-byte cells.

### The result

| | Binary (Caltech, VG×siglip) | Region (VG×dinov3_patch) |
|---|---|---|
| H1: any K beats 2 | **No** | Yes — K ∈ {6,8,12,16} |
| **H3 recommendation** | **keep 2** | K=6 (2.68× cost) |
| Best ignoring cost | K=1 | K=16 (−0.0075) |
| **H4: benefit is `rule_inefficiency`** | ok (moot) | **FAILS** |
| A/B `screen_agrees` | true | true (p=0.29) |

**Binary: more folds actively hurt.** Deep-window (≤200 votes) regret climbs
monotonically from 0.00125 at K=1 to 0.00426 at K=16. Production's 2 is safe;
K=1 is arguably better.

**Region: H1 passes, but the pre-registered mechanism is inverted.** The plan
predicted extra folds cut estimator variance (`rule_inefficiency`) against a
calibration→test shift term it stated K "cannot touch at all". Measured, it is
exactly backwards — `rule_inefficiency` gets *worse* with K (+0.0124 at K=6,
+0.0175 at K=16) and `calibration_shift` supplies the entire benefit (−0.0191,
−0.0260). The plan's own H4 clause then applies: *the recommendation should not
be trusted even if H1 passes*. **K=6 is a measured effect with a wrong
explanation, not a ship candidate.**

**The A/B validates the screen but shrinks the effect.** Live region K=6
`d_regret` is −0.0035 against the screen's −0.0050, p=0.29 — no significant
disagreement, so the nested-prefix screen is a sound proxy and acquisition
feedback does not overturn it. But the live point estimate sits **below the
pre-registered MARGIN of 0.005**, so K=6 does not clear the ship bar on the live
number either.

Worth weighing against any future adoption: at K=6 region, `cal_share` is
**0.76** (0.90 at K=16) — three-quarters of every step is calibration the user
waits through after each vote. The 4× cost ceiling looks too generous for an
interactive path.

### Recommendation

**Ship nothing; `calibrate_count` stays 2 in both modes.** Binary says so
outright. Region's candidate fails its own mechanism check and lands under the
margin once acquisition feedback is allowed to move. The follow-up worth having
is not a bigger fold sweep but an explanation of *why* K moves the shift term —
that premise being wrong is the actual finding here.

### Two more fixes in this PR

- `launch_all.sh` passed `--gres=none` through literally and hardcoded
  `--partition=gpu`; this cluster's submit filter rewrites that and rejects the
  job, `--parsable` returns an empty id, and **both A/B arms silently failed to
  launch**. Latent for every study using the full chain with `CALIB_GRES=none`.
  Fixed, plus `require_jobid` after every `sbatch`.
- The A/B's own **control** arm (`CALIB_FOLD_COUNTS="2,$K"` collapses to one
  value at K=2) crashed its per-arm analyze with `KeyError: 'voting'`. Nothing
  was lost, but a control arm should report "nothing to contrast", not a
  traceback.

Both logged in `scripts/experiments/LESSONS.md` under one heading, because they
are the same mistake: **a signal that can be satisfied by something other than
the thing you are waiting for.** I made the same error a third time while
monitoring — treating an existing `STATUS.md` as proof of completion when it had
been written by an earlier stage.

Artifacts: `REPORT.md` (with the `screen_agrees` table), `REPORT_screen_only.md`
(pre-A/B), `summary.json`, `STATUS.md`, under
`/home/hltcoe/sgreenberg/experiments/calibration-folds-2897/`.

Issue #2905, opened 2026-08-07
Should the acquisition offset be gated by voting mode? Run the region-voting check that #2877 did not


samggreenberg:
**Difficulty:** medium — the harness knob, the launcher and the analyzer all exist and have now run twice. What is new is the *cost* of the environment and one unmeasured decision at the end.

**Recommended Claude model: Opus 5.** Not for the mechanics, which are bounded, but because the output is a shipping decision about a constant that is already live for every user, and the last attempt at this check got the environment wrong in a way that looked completely healthy.

## What

Decide whether `ACQUISITION_INCLUSION_OFFSET` should be gated by voting mode, and run the measurement that decision needs.

Two environments have now measured the offset and they disagree:

| environment | voting mode | verdict on `-3` |
|---|---|---|
| `coco_val × siglip2` (#2876) | binary | ships: positives 4→18, cost **−0.011** |
| `visual_genome_m × siglip` (#2877, PR #2891) | **also binary** | **fails**: cost CI **[+0.003, +0.022]** vs a +0.01 tolerance; only `k=-1` passes |

**Both are binary voting.** #2877 pre-registered its arm as *region* voting and it is not — `visual_genome_m__siglip` carries no `patch_grid`, so `region_voting=True` silently fell back to whole-image training, whole-image scoring and the `cap50` (binary) blend schedule. See the correction banner on PR #2891 and `docs/experiments/acquisition-inclusion/REPORT_SECOND_ENVIRONMENT.md`.

So the gating question is **wide open in both directions**, and the shipped `-3` currently rests on a single environment that a second environment contradicts.

## Why this is worth doing properly

**The disagreement already has a mechanism, and it predicts where region voting lands.** In the second environment, `regret` (cost − oracle cost) is *flat* in every negative-`k` arm — the cut estimator is blameless — while **AP rises** (0.349 → 0.371, p<1e-5) and **oracle cost rises too** (0.395 → 0.410). Aggressive acquisition sharpens the head of the ranking and blurs its tail; AP sees the first, a globally-placed reporting cut sees the second. COCO escaped this only because it was starved hard enough that any positive helped everywhere.

[...]

PR #2909, opened 2026-08-08
Acquisition offset #2905: the region-voting environment says don't gate by voting mode — ship -1

samggreenberg:
**Answers #2905.** The third environment for `ACQUISITION_INCLUSION_OFFSET`, and the first that actually region-votes.

3864/3864 cells, **0 failures, 0 zero-byte**, ~5h, zero GPU. Report: [`docs/experiments/acquisition-inclusion/REPORT_REGION_VOTING.md`](docs/experiments/acquisition-inclusion/REPORT_REGION_VOTING.md).

## Verdict: don't build the gate. Ship `-1`, revert `-3`.

`-3` **passes** its pre-registered ship rule under region voting — positives 14 → 20 per 100 votes, final-cost CI [−0.0034, +0.0095] against a +0.01 tolerance, spikes flat. Taken alone that reads as a third endorsement. It shouldn't be:

**1. The three environments split by environment, not by mode.**

| environment | voting | verdict on `-3` |
|---|---|---|
| `coco_val × siglip2` (#2876) | binary | **ships** — cost −0.011, AP 0.696 → 0.817 |
| `visual_genome_m × siglip` (#2877) | binary | **fails** — cost CI [+0.003, +0.022] |
| `visual_genome_m × dinov3_patch` (this) | **region** | passes |

The largest disagreement is *within* binary voting, which no mode gate can reach. **`-1` is the only value passing in all three.**

**2. The mechanism `-3` was sold on is absent here.** #2877 and this run share the dataset, the **same 23 categories** and the same 24 seeds, so they pair cell-for-cell — the only controlled voting-mode contrast available for this knob. Response to `k=-3` on 536 shared cells:

| endpoint | binary | region | region − binary |
|---|---:|---:|---:|
| **AP (head)** | **+0.0283** | +0.0003 | **−0.0281** [−0.0361, −0.0202] |
| **oracle cost (tail)** | +0.0084 | +0.0039 | −0.0047 [−0.0122, +0.0028] |

> The tail-blurring is mode-invariant. The head-sharpening is binary-only.

Region voting pays the bill and receives none of the goods. Not a dead metric: the falsifier moves AP −0.0153 (p<1e-5) in the same cells.

## The premise, verified before the run

`patch_grid` on **4193/4193** medias (`siglip`, what #2877 measured: **0/4193**). `preflight.sh --require-region-voting DATASET:EMBEDDER` now opens the pickle and refuses to submit without it. No fallback warning fired in 3864 cells.

While adding it I found the arm-collision and zero-byte checks only ever scanned `results-ab/` — **silent no-ops for every study whose arms live under `results/`**, which is all of the acquisition and anchor sweeps, including #2877. A gate reporting `ok` without having looked is worse than a missing one.

## The cost of this recommendation, stated plainly

On a starved COCO-like environment `-1` finds **6** positives per 100 votes where `-3` finds **18**. That's the strongest case against this change. It's outweighed because `-3` rests on one environment, a second rejects it, and its justifying mechanism is absent in the third.

**The successor is a supply-dependent offset, not a mode gate.** Under binary voting the benefit concentrates sharply in *starved* cells and turns negative in well-supplied ones — AP response slope **−0.0207** on log prevalence (CI [−0.0259, −0.0159]), **−0.0402** on a leave-one-out baseline. The detector knows its own positive count at run time. That subsumes the mode question, since mode only ever entered through label supply.

## If the gate is built anyway, the issue's implementation is wrong

`ctx.embedder_type` is the detector's **locked** type and is `""` for a legacy detector; `_patch_embedder_for_snap` reads the **dataset's** patch slot, which is what `_blend_schedule_for_snap` already gates on. A legacy detector on a patch dataset would get `slow_cap50` from the schedule and *binary* from the offset — reintroducing the exact inconsistency #2905 exists to close. The right place is `_fused_threshold`, which already holds `clips_dict` when it parks the estimator on `det_ctx.anchored_cut_cache` (`vtscore/detectors/training.py:201`).

## A correction I had to make to my own analysis

The unifying "starvation remedy" story appeared in **both** modes when cells were binned on how many positives `prod` found — but the response is measured against that same `prod` run, so mean reversion manufactures the curve. Re-cut on arm-independent axes (category prevalence; a leave-one-out baseline), the **binary** curve survived at full strength and the **region** curve vanished — significant on the contaminated axis, null on both clean ones.

That reversed the conclusion: at *matched* prevalence the modes still differ (most-starved bin: binary +0.0391, region +0.0091 n.s.), so the mode difference is real and **not** reducible to supply. Both the contaminated and corrected analyses are kept beside the experiment.

## Changes

- `ACQUISITION_INCLUSION_OFFSET` `-3` → `-1`, with the three-environment history and the successor design in its docstring.
- `docs/ML.md` updated; the `PRODUCTION_SCHEDULE_BY_MODE` asymmetry recorded as **measured**, not an oversight — the schedule was measured to differ by mode, the offset measured not to.
- `test_acq_inclusion.py` pins the new constant and moves its "did the offset go absolute" probe to reporting inclusion −1, where an absolute reading would now collapse. **24 tests pass.**
- Two `LESSONS.md` entries: the vacuous preflight pass, and the mean-reversion trap.
- Seeds declared at 48 and 24 run, so a top-up is `--seeds 24-47` with every cell keeping its index. Not needed — paired SD **0.0709** gives ±0.0060, tighter than either prior environment.

Includes a cherry-pick of #2906's pre-#2886 `RegionVector` pickle shim; without it this arm dies on cell 1. Drop it if #2906 merges first.


samggreenberg:
**Test status.** 104 tests run across `test_acq_inclusion.py`, `test_acquisition_threshold.py`, `test_fold_count_variant_rows.py` and `tests/sorting/test_sorting.py`: **103 passed, 1 failed**.

The failure is `test_reported_xcal_seconds_is_billed_for_the_live_count`, and it is **not from this branch** — it fails 5/5 on a clean checkout of `dev` (`00664df5`) with none of these changes applied. It arrived with #2897 / PR #2906, merged earlier today. Filed as **#2911** with a diagnosis: `fold_seconds` times the fold fits while `xcal_seconds` times the whole cross-calibration step, and on the tiny test fixture the latter's fixed overhead dominates. On real data the assertion holds by ~18× (0.088 vs 1.550), and `fold_seconds` scales exactly linearly in K, so the instrumentation is fine.

The 24 tests that directly pin `ACQUISITION_INCLUSION_OFFSET` all pass. Note `test_sorting.py`'s `acq_threshold >= threshold` assertion is satisfied by any negative offset, so it does not constrain the −3 → −1 change either way.

Also rebased onto current `dev`: #2906 merged and brought the `RegionVector` pickle shim I had cherry-picked, so that commit was correctly dropped as already-applied. The only conflict was `LESSONS.md`, an append/append — both sides kept.

Issue #3160, opened 2026-08-17
gres/gpu:v100 is two different devices, and one changes siglip2_l fp32 by 1.5e-4


samggreenberg:
**Recommended Claude model: Opus 4.8.** Subtle numerics plus a scheduling/provenance change that touches every future pile rebuild; a wrong-but-plausible answer here silently invalidates cross-cell comparisons.

#3144 closed with: *"No experiment needed. Same code, same fp32 math. Cross-GPU fp32 drift is ~1e-7 — far below anything the studies resolve. This one is safe to land on its own."*

The #3143 precision study measured that claim as a control, and it does not hold. **`gres/gpu:v100` is not one device**, and one of the parts it covers changes `siglip2_l` fp32 vectors by **1.5e-04**.

### Measured

Same commit, same 4193 VG images, fp32 throughout. Median 1−cos, `visual_genome_m × siglip2_l`:

| node | GPU as torch reports it | vs published pile cell | vs L40S | `siglip2_l` build |
|---|---|---|---|---:|
| rack7n03 | Tesla V100S-PCIE-32GB | **0.0 (bit-identical)** | 2.7e-12 | 380 s |
| rack5n03 | Tesla V100-SXM2-32GB-LS | 1.5e-04 | 1.5e-04 | 442 s |
| — | NVIDIA L40S | 2.7e-12 | – | 177 s |

Three of the four devices agree to ~1e-12. The **V100-SXM2-32GB-LS on rack5n03** is the outlier, at 1.5e-04 — 30× the 0.005 margin the calibration studies resolve, and ~50× the drift that switching the whole forward to fp16 costs (2.9e-06). On `siglip` the same two nodes agree to 7.6e-13, so it is specific to the SO400M/384 geometry.

Ruled out, each by a bit-identical diagnostic arm:

- **TF32** — `fp32` with `cudnn.allow_tf32=False` is bit-identical to default fp32 on the L40S (median exactly 0.0). TF32 was never active; `matmul.allow_tf32` defaults to `False`.
- **cuDNN algorithm nondeterminism** — `cudnn.deterministic=True` is bit-identical to default on *both* nodes (exactly 0). Each node is perfectly self-consistent across runs and flag changes.

Remaining candidates, untested: a capability- or SM-count-selected attention/SDPA backend, or GEMM tiling and accumulation order that differ with SM count.

### Why it matters

[...]

PR #3171, opened 2026-08-18
#3146: the fast image processor was already on — name the backend, measure what is left

### LINKED ISSUE #3146: Image processors load slow (PIL) path — becomes the top bottleneck once decode and precision are fixed
Every image embedder builds its processor with no `use_fast` argument, e.g. `vtscore/media/image/embedder_siglip2_l.py:91`:

```python
self._processor = load_pretrained_local_first(
    AutoProcessor.from_pretrained, SIGLIP2_L_MODEL_ID, cache_dir=cache_dir, token=hf_token()
)
```

so we get the slow PIL/numpy resize + normalize path, single-threaded, on the calling thread inside `_forward_pil_batch`. The one place in the codebase that mentions the flag sets it to `False` explicitly (`vtscore/media/video/embedder_xclip.py:106`).

### Evidence

384 real VG images (jobs `507149` / `507150`). The processor is a near-constant ~2.0 s on V100 and ~0.9 s on L40S regardless of model, batch size or precision — it does not scale with anything we tune elsewhere. That makes it small today and dominant tomorrow:

| config | processor share, `siglip` |
|---|---:|
| V100 fp32 batch 32 (today) | 28% |
| V100 fp16 batch 128 + 8 decode workers | **68%** |
| L40S fp16 batch 128 + 8 decode workers | **66%** |

So once #3143 and #3145 land, this is the top remaining cost for the whole-image encoders.

### Fix

Pass `use_fast=True` (torchvision-backed), or move resize/normalize onto the GPU. Check whether xclip's explicit `use_fast=False` is a deliberate correctness workaround before changing it — it may be load-bearing.

### This needs an experiment before adopting — `experiment`

The fast processor is not a drop-in numeric equivalent: torchvision's resize differs from PIL's in interpolation and antialiasing, so the model sees different pixels. This is a larger perturbation than fp16, not a smaller one.

Required before flipping the default:

1. Rebuild one pile cell both ways; report per-media cosine drift and retrieval rank correlation.
2. Re-run one benchmark arm both ways against the study margins.
3. Confirm the fast path handles the inputs the slow one does — CMYK, palette, EXIF-rotated, grayscale — since `decode_bounded_rgb` hands it whatever the corpus contains.

Same all-or-nothing constraint as #3143: the pile is a shared cache and must not be partially rebuilt. Sequence this **after** #3143 so a single rebuild absorbs both perturbations and only one drift study is needed.

Context: docs/experiments/overview-bench/REPORT.md, "Cost is two different things".

samggreenberg:
Experiment for #3146, plus the two knobs it turned out we needed.

## The premise does not survive contact with the installed library

#3146 reads the source — every image embedder builds its processor with no
`use_fast` — and concludes they are all on the slow PIL path. The source does say
that. Under the installed `transformers` 5.12.1 it is false: **v5 removed the
`Fast` suffix**, so `SiglipImageProcessor` *is* the torchvision implementation
and the slow one was renamed `SiglipImageProcessorPil`. Passing nothing already
selects torchvision, and `use_fast` is deprecated in favour of `backend=`.

That is settled empirically rather than by reading a class name: the torchvision
rebuild reproduces the **published pile cell** to 7.6e-13 (`siglip`) and 2.7e-12
(`siglip2_l`), while the PIL rebuild sits at 2.4e-06 and 1.5e-04. The pile is
torchvision-built. The flip #3146 proposes already happened, silently, through a
`transformers>=4.49` pin that spans the version where the default changed.

## What the study measured instead

| | |
|---|---|
| **GPU preprocessing** (the issue's *other* proposed fix) | 1.68× ± 0.04 on `siglip`'s embed path — but **1.09× per pile cell**, and ~1.02× for `siglip2_l` |
| **The speedup we already have** | `pil_cpu` is 0.52×, so the silent torchvision default is a 1.9× embed-path win nobody attributed to anything |
| **Vector drift** | the models invert: PIL costs `siglip2_l` 1.5e-04 and `siglip` 2.4e-06; CUDA costs `siglip` 1.2e-04 and `siglip2_l` 5.4e-08 |
| **User-visible** | Spearman ρ = 1 everywhere, but `tv_cuda` changes the **top-1 result on 5% of `siglip` text queries** |
| **Benchmark** | cannot resolve at 96 cells (needs ~1100); the arms choose the threshold identically on **99.9%** of 13,627 paired steps. A 64-seed power run is in flight |
| **Awkward inputs** | the checklist inverts — on the corpus path every backend is identical on all eleven modes, and on the direct path **PIL is the more fragile one** (raises on 8 modes to torchvision's 4) |

## The finding that outlives the issue

**The backend is a second unrecorded axis on the shared pile**, alongside the
device axis #3160 is chasing, and a larger one. Two hosts resolving different
`transformers` produce different vectors from identical code and weights, with
nothing recording which built any cell.

It is *separable* from #3160's CPU-dispatch axis, which had to be checked because
both produce exactly one 8-bit level (7.843e-03) at the pixel level and are
therefore indistinguishable by magnitude. The dispatch matrix separates them: the
backend moves 53–59% of elements against dispatch's 13.7% and survives both
settings. It also explains the model inversion above — **CUDA agrees with the
AVX-512 CPU kernel at 384px (0.03% of elements) and disagrees at 224px (11.48%)**.

## Changes

- `vtscore/config.py`: `VTSEARCH_IMAGE_PROCESSOR_BACKEND` and
  `..._DEVICE`. Both default to `auto`, which passes nothing and is
  byte-for-byte what built the pile.
- `embedder_siglip.py` stops naming `SiglipImageProcessor` directly. A concrete
  class pins the code to whatever that identifier currently means — the exact
  thing that changed — and it cannot honour a backend request at all, so the
  knob would have been inert there alone.
- `scripts/experiments/fastproc/`: the study. Every arm asserts the processor
  class it *actually* loaded and refuses to build a cell when it contradicts the
  arm table, because transformers **warns and continues** when a backend is
  unavailable — the default outcome of an impossible request is a mislabelled arm.
- `tests_lib/core/test_image_processor_backend.py`: 18 tests, mostly about
  silence — that the default passes nothing, that a typo degrades rather than
  quietly changing what gets embedded, and that the `backend=`/`use_fast=`
  spelling is resolved from the installed version rather than assumed.
- `docs/experiments/fastproc-3146/` + a generated `docs/reports/` reading copy.

## Recommendation this PR does *not* implement

Change the backend default from `auto` to `torchvision`. It is the one change
worth making regardless of how the power run lands — naming the backend costs
nothing if the difference does not matter and is essential if it does — but it
is a behaviour change for hosts on `transformers` 4.x and belongs in its own PR
with its own review.

## Verification

`./run-tests.sh` on the GRID: 8696 passed, 56 skipped, 1 xfailed. The first run
failed `pyright` only (a `dict[str, object]` splat poisoning six call sites);
fixed and re-running as job `516041`.

Closes #3146

The issue's experiment requirement is discharged in full: pile cells rebuilt
both ways with drift and rank correlation, a benchmark arm run both ways at 96
and then 1013 paired cells, and the CMYK/palette/EXIF/grayscale checklist. Its
*proposed fix* is refuted as a no-op. The deferred recommendations are now their
own issues — #3172 (EXIF orientation) and #3173 (name the backend default) — so
nothing is still owed on #3146 itself.


samggreenberg:
**Power run landed, and it changes the recommendation from "not yet" to "no".**

1013 paired cells per arm (64 seeds, 10.7× the first grid). The extra cells worked as intended — SE on `cost` fell 0.0087 → **0.0023**, a factor of 3.8 against the √10.7 = 3.3 the cell count predicts — but the result is not a flat null:

| metric | paired diff ± SE | verdict |
|---|---:|---|
| **regret** | **−1.1e-04 ± 0.0020** | **below margin (resolved)** |
| cost | 0.0027 ± 0.0023 | cannot resolve |
| average precision | −0.0038 ± 0.0021 | cannot resolve |
| fnr | 0.0047 ± 0.0027 | cannot resolve |

`below margin` requires `|diff| + 2·SE < 0.005`. The standard errors are now small enough; the **point estimates** are not near zero, so `cost` and `fnr` carry upper bounds of 0.0073 and 0.0101 — the latter twice the margin.

So ten times the cells moved exactly one metric across the line, and the useful reading is that the effect is *probably real and probably near the margin* rather than absent. That is the regime where adopting on a null is a mistake. Against 1.09× per pile cell, a 5% top-1 change on `siglip` text queries, and a full pile rebuild, the study now recommends **not** adopting GPU preprocessing.

Coverage: 1013 usable of 1024 per arm. Eleven cells produced zero rows in both arms identically — six are `siglip` × `ball` at different seeds, plus `cat` ×2 and `sink`. Near-degenerate categories, counted rather than dropped.

Also: `./run-tests.sh` on the GRID now passes clean — **8696 passed, 56 skipped, 1 xfailed, RUN PASSED**, all gates green including the pyright fix.

Report §6 and the recommendations are updated, and the HTML reading copy is regenerated.

Issue #3144, opened 2026-08-17
Pile builds default to v100, the slowest GPU on the cluster


samggreenberg:
`scripts/experiments/pile/launch_pile.sh:17` hardcodes the fallback:

```sh
GPU_TYPE="${VTS_GPU:-v100}"
```

Every pile cell built so far was embedded on a V100 (confirmed via `sacct` on jobs `495245`/`495246`/`495266`/`495377`/`495378`/`495379` — all `gres/gpu:v100=1`). L40S, A100 and H100 nodes were idle at the time and were idle again when this was checked on 2026-08-17.

### Evidence

Same 384 VG images, same code, fp32, batch 32 (jobs `507149` / `507150`):

| | V100 | L40S | speedup |
|---|---:|---:|---:|
| `siglip` | 52.5 img/s | 87.6 img/s | 1.7x |
| `siglip2_l` | 14.0 img/s | 32.3 img/s | **2.3x** |

For 12,000 `siglip2_l` images that is 855 s -> 372 s for a one-word change.

### Fix

Default to a faster type, or restore the `scontrol`-based auto-pick that used to live in the `vtsearch` launcher (see the identities note — it queried free GPUs and requested the most-available type, then got overwritten when the repo script was re-copied). Note this cluster rejects untyped `--gres=gpu:1`, so "any" has to resolve to a real type.

Worth pairing with the queue reality: pinning `l40s` once caused ~5-day waits when only two l40s nodes existed. rack4n01's 8 GPUs are back up as of 2026-08-17, but an auto-pick is more robust than any new pin.

### No experiment needed

Same code, same fp32 math. Cross-GPU fp32 drift is ~1e-7 — far below anything the studies resolve. This one is safe to land on its own.

samggreenberg:
Addressed in #3150 — restored as an auto-pick rather than a new default, since `vtsearch-slurm.sh` already showed how the opposite pin fails.

`scripts/slurm/pick_gpu.py` reads `scontrol show node` and requests the **fastest candidate type that has enough free GPUs to start on now**, degrading to most-free, then largest-pool, then a fallback. Preferring fastest-that-can-start over "most-available" matters here: pure max-free would have re-elected v100, since the v100 pool is the biggest. `h100`/`h200` stay out of the default candidate list because `4gpu_tier` caps them at 0. `VTS_GPU` still overrides outright.

[...]

Issue #3143, opened 2026-08-17
Image embedders run fp32 with no autocast — 4.2x left on the table for siglip2_l


samggreenberg:
Every image embedder loads its model in fp32 and runs the forward with no autocast and no `dtype` argument — `vtscore/media/image/_cross_modal_shared.py:117`, and the same shape in `_dinov3_shared.py` / `_dinov2_shared.py` / `_eupe_shared.py`. On a V100 that means 15.7 TFLOPS of fp32 instead of 125 TFLOPS on the fp16 tensor cores.

### Evidence

Measured 2026-08-17 on 384 real VG images through the actual `vtscore` decode + processor + forward path (GRID jobs `507149` V100, `507150` L40S). The fp32 baseline reproduces the pile build almost exactly — 855 s projected for 12k `siglip2_l` images against the 830 s the build actually logged — so these transfer.

GPU forward time, 384 images:

| | V100 fp32 | V100 fp16 | speedup |
|---|---:|---:|---:|
| `siglip` | 1.3 s | 0.4 s | 3.3x |
| `siglip2_l` | 22.1 s | 5.2 s | **4.2x** |

End to end for 12,000 `siglip2_l` images: 855 s -> 329 s on the same V100, from precision alone.

### Fix

Load in fp16/bf16 (or wrap the forward in `torch.autocast`), picking by device capability — V100 is sm_70, so fp16 only, no bf16. Keep an env escape hatch to force fp32.

### This needs an experiment before adopting — `experiment`

fp16 changes the vectors. Cosine similarities shift by roughly 1e-3, and the calibration studies resolve effects of 0.005, so this cannot be turned on silently.

Required before flipping the default:

1. Rebuild one pile cell (suggest `visual_genome_m x siglip2_l`, the cheapest non-saturated one) in both precisions.
2. Report the distribution of cosine drift per media, and rank correlation of the retrieval ordering.
3. Re-run one benchmark arm both ways and confirm cost/AP move by less than the margin the studies care about.

**Do not partially rebuild the pile.** A pile with some cells fp16 and some fp32 is a confound that would surface months later as an unexplained arm difference. It is all cells or none, and the existing published results (#3129, the vg_box re-run) are against the fp32 pile.

Context: docs/experiments/overview-bench/REPORT.md, "Cost is two different things".

samggreenberg:
[...]

PR #3178, opened 2026-08-18
#3160: a type is not a device — and the device was never the problem

### LINKED ISSUE #3160: gres/gpu:v100 is two different devices, and one changes siglip2_l fp32 by 1.5e-4

samggreenberg:
Closes #3160.

#3143 §5 measured two nodes both answering to `gres/gpu:v100` producing `siglip2_l`
vectors **1.5e-04** apart and left the cause open, with two candidates: a
capability-selected attention backend, or GEMM tiling that follows SM count.

**Both are wrong, and the GPU is not involved.** Fed the same preprocessed
pixels, the two V100 parts compute `siglip2_l` bit-identically — all 27 vision
blocks and the image features. The divergence enters on the **host**, in the
384px resize: PyTorch dispatches a different CPU kernel under AVX2 than under
AVX-512, and the two disagree on 12.3% of pixels, each by exactly one 8-bit
level (2/255). The nine "outlier" nodes are exactly the nine whose CPUs have no
AVX-512.

Forcing `ATEN_CPU_CAPABILITY=avx2` on an AVX-512 host reproduces the AVX2 host's
pixels **exactly**, and pinning it on both hosts takes `siglip2_l` from median
1−cos 1.3e-04 (0% of rows bit-identical) to **median 0, max 8.9e-16, 76%
bit-identical** — while making the resize **26% faster** (2.18 ± 0.02 s →
1.62 ± 0.02 s per 256 images).

So **#3144's premise was right about the GPU** (cross-part disagreement with the
input pinned is ~1e-15) and the auto-pick is not the defect. What is unsafe is an
unpinned *host*.

## What ships

- **`launch_pile.sh` pins the dispatch**: `ATEN_CPU_CAPABILITY=${VTS_CPU_CAPABILITY:-avx2}`.
- **Per-cell provenance sidecars** (`build_pile.py`): GPU name/capability/SM
  count/driver, **CPU model and resolved CPU capability**, torch/cuDNN,
  transformers version and resolved processor classes, precision, node, SLURM
  job, commit — plus a `vectors_sha256` fingerprint that outlives the cell.
  `--provenance` prints the table and flags a pile that mixes build
  environments; `--backfill-provenance` stamps the 21 existing cells and
  recovers their build node from `sacct` (as `hostname_recovered`, never as
  `hostname`).
- **`VTS_GPU_NODE`** pins `--nodelist` and derives the gres type from the node,
  for an exact rebuild.
- **Study code** under `scripts/experiments/gpu_node/` and the report under
  `docs/experiments/gpu-node-3160/`.

## Found on the way

`vg_box_small`'s cells were built on **rack8n06 (AVX2-only)** while
`vg_box_medium` and `vg_box_large` were built on rack7n03 (AVX-512) — so the
three bands #3129/#3156 compare were not preprocessed identically, for their
`siglip2_l` cells. `siglip` and `dinov3_patch` are 224px and bit-identical
across the split, so the region-voting arm of those studies is clean.

## Does the drift reach a decision?

A paired benchmark (production defaults; only the gallery's build host differs)
ran 4096 array tasks, all COMPLETED, zero zero-byte cells → 2027 paired cells,
294,970 paired steps. Half the grid is a **placebo**: `siglip`, whose two piles
are identical, and which duly returns approximately exact zeros (SEs ~25×
tighter than the treated arm, `cost` bit-identical on 99.2% of steps).

On the treated embedder, no metric is distinguishable from zero after accounting
for testing seven, and every point estimate is ≤0.006. `rule_inefficiency` at
−0.0057 ± 0.0022 (2.6 SE) was flagged rather than claimed, with the reading
pre-registered before the number existed.

**The replication settled it.** An independent grid at 2× the treated cell count
(2033 paired cells, 296,166 paired steps, zero zero-byte cells) puts
`rule_inefficiency` at **−0.0015 ± 0.0015** — same sign, a quarter the magnitude,
1.0 SE from zero and resolvably below the margin. It was multiplicity, the same
shape #3143's `n_good` had. **Five of seven metrics now resolve below 0.005 on
the treated embedder alone**, and three of the seven signs flipped between the
two grids. Nothing was pooled.

So the drift is a **reproducibility defect, not a correctness one** — answered on
4060 paired cells across two independent grids.

Note the pooled rows in the analyzer's table read tighter than the truth here
because the pool is half placebo; the report says so and uses the per-embedder
split.

## Tests

`./run-tests.sh` on the GRID (job `516026`): **ALL 8678 TESTS PASSED** (56
skipped, 2 xfailed), all gates green.


samggreenberg:
FYI: this PR's work was deleted from `dev` by commit `2bb2f054b` (merged via #3184) 23 minutes after it merged, along with #3182's and #3175's. Restored in #3206; the recurring pattern is filed as #3205.

Issue #2881, opened 2026-08-07
Measure the one-constant EVT tail-α cut (α ≈ 0.16 on the Gumbel low component) against the fold-anchored threshold


samggreenberg:
Medium: one new cut rule (closed-form, ~30 lines), then a Grid cell run and a paired analysis against the current production threshold. The design judgment is the hard part, not the code.

**Recommended Claude model: Opus 4.8.** The plumbing is Sonnet-shaped, but this family has now been mis-sized *twice* by the theory bench in the same direction (see #2846 / `REMEASURE-2846.md`), and the incumbent baseline moved mid-study once already. A wrong-but-plausible read here is exactly the failure mode that has cost this line two runs.

## What this is

`docs/experiments/gmm-cut/REMEASURE-2846.md` closed the Gumbel **crossing** work — the six `gumbel_*` / `gumbel_any_*` rules do not clear the plain Gaussian `pooled_priorfree`, let alone production. But it discharged `REPORT.md`'s pre-registered recomputation and the tail-level finding **survived**, over 511 cells with the previously-discarded swapped fits now kept:

| tail model | median α (#2836 → now) | IQR ratio (#2836 → now) | stable (bar: < 3) |
|---|---|---|---|
| Gaussian low component | 0.069 → 0.070 | 4.93 → 5.54 | no |
| Gumbel low component | 0.165 → **0.158** | 2.22 → **2.38** | **yes** |

The report's words: *"cut the Bad tail at α ≈ 0.16" is still a coherent one-constant rule that needs no crossing at all — which is now the only part of the EVT work with a future.*

Nobody has measured it as a rule. `oracle_lo_sf_evt` is currently only recorded **diagnostically** — the survival of the fitted low component evaluated at the *oracle* cut (`vtscore/eval/cut_rules.py:383`). The finding is that this number is stable across cells; the rule inverts it.

## Why it is a different animal from the crossing rules

- **No orientation question.** It reads only `GumbelNormalFit1D.lo_survival` (`vtscore/training/evt_mixture.py:309`), which already returns whichever component is the low one. The entire `modes_swapped` / `allow_swapped` axis that #2836 and #2846 spent themselves on is irrelevant here.
[...]

Issue #3187, opened 2026-08-19
Re-run `analyze_cut.py` over the gmm-cut cells and publish the corrected decomposition tables


samggreenberg:
**Difficulty: low, but gated on cluster data.** **Recommended Claude model: Sonnet 5.** The analysis is a re-run of an existing script over existing outputs; the judgment needed is reading the new integrity columns and deciding whether any verdict moved.

## What happened

`analyze_cut.py`'s two decomposition tables averaged each term with a per-column `mean()`, which skips NaN independently per column. The oracle variants (`pooled_supervised`, `pooled_sim_oracle`) do not fall back — a step whose oracle cut is not finite emits no row at all — so those steps pivot to NaN and were dropped from the three oracle-dependent terms (identification, misspecification, transfer) while still counting toward `total`.

The tables show it: the four terms sum to **+0.0598** (#2836) and **+0.0597** (#2879's re-measure) against a stated total of **+0.0623** in both. The `residual` column was meant to catch exactly this and could not — a NaN link makes the row's residual NaN too, so `.mean()` skipped it and the guard read `0.0` while rows were being dropped.

Fixed in the analyzer (this issue's parent PR): both tables now aggregate over complete rows only, so every term and `total` cover the same steps; `n_incomplete` counts the drops, `cost_residual` pins the cost-unit chain, and `summary_cut.json` carries both next to `dominant_error_term`.

## What is still owed

The corrected *numbers*, which need the cluster data this session could not reach.

1. Re-run `analyze_cut.py` over the existing `results/cells/task_*.csv` + `task_*__cutdiag.csv` for both studies. **This is a re-analysis, not a re-sweep** — no GPU time, no new cells.
   - #2836: the run behind `docs/experiments/gmm-cut/REPORT.md`.
   - #2879: experiment `/exp/$USER/calibration-cut2846`, behind `docs/experiments/gmm-cut/REMEASURE-2846.md`.
2. Replace the annotated tables in both documents with the corrected terms, and delete the annotation blocks.
3. Report `n_incomplete` alongside — how many steps lacked an oracle link is itself a fact about the study nobody has ever seen, and it bears on how well-measured the three oracle-dependent terms are.
4. Confirm the verdicts are unchanged. Expected, not assumed: `transfer` dominates at 65–68% either way, and `dominant_error_term` is read from the same means. If `dominant_error_term` moves, that is a finding, not a formatting fix.

## Why it probably does not change any conclusion

[...]

Issue #3116, opened 2026-08-12
The regret decomposition's reference point moves with K, so #2897's H4 could not have worked


samggreenberg:
**Recommended Claude model: Opus 4.8.** The failure mode here is a wrong-but-plausible statistical reading; the code change is small.

Refs #2897

## The defect

`_operating_metrics` (`vtscore/eval/voting_iterations.py:655-661`) splits regret as:

```python
c_thr             = oracle_cut(cal_scores, cal_labels)    # best cut ON the pooled calibration set
cal_oracle_cost   = operating_cost(test_scores, test_labels, c_thr)
rule_inefficiency = cost - cal_oracle_cost
calibration_shift = cal_oracle_cost - o_cost
```

`c_thr` — the reference cut that *defines* where the split falls — is estimated from the pooled calibration set, and **that set grows linearly in K**. So when a study sweeps the fold count, it is moving the yardstick it is measuring against.

Mechanically, as K rises `c_thr` converges on the test-oracle cut, which shrinks `calibration_shift` and widens `rule_inefficiency` **from one cause, in opposite directions**, with the sum pinned to regret by construction. #2897 observed exactly that and reported it as "the benefit landed on the shift term, so the variance story is wrong."

That reading does not hold. `rule_inefficiency` is not a variance — it is a signed cost gap between two specific cuts — and it was **negative in every row of the run** (binary `le_20`: −0.291 at K=1 → −0.080 at K=16), rising *toward* zero. A negative value means the trained conformal cut beats the calibration-set "oracle" on test, i.e. the oracle is overfitting a handful of scores. Nothing in the study says variance rose.

The plan's premise — *"a bias term (calibration→test shift) that K cannot touch at all"* — is false by construction, since `calibration_shift` is defined through an estimate taken from the K-dependent calibration set.

## Verified

The identity `rule_inefficiency + calibration_shift == regret` holds exactly on the shipped numbers (binary `le_20` K=1: `−0.291185 + 0.355423 = 0.064238` vs reported regret `0.064237`; region `le_100` K=6: `−0.028873 + 0.150901 = 0.122028`, reported `0.122028`). The anti-correlation is algebraic, not empirical.

## What to change

- **Re-decompose against a reference that does not move with the arm** — a large held-out calibration oracle, or `c_thr` frozen at the baseline K across all arms. Then bias and variance are identifiable and H4 becomes answerable.
[...]

PR #3248, opened 2026-08-24
#2883: `transfer` is a variance measured against an optimistic reference — not a cost

### LINKED ISSUE #2883: Aim the next calibration study at the fit, not the cut: `transfer` is +0.041 and the cut axis has no headroom left
**Recommended Claude model: Opus 4.8.** Study design plus mixture-estimation numerics, where a wrong-but-plausible design produces a clean table that means nothing — which is exactly how the last two runs on this axis went.

> **Does not supersede #2881.** #2881 measures the one-constant EVT tail-α cut, which is still a *cut* of the same unanchored fit — so at first glance it points the other way from this issue. It doesn't: #2881 is one cheap, pre-registered run that closes the cut axis honestly (it is the only EVT variant whose fallback rate is not what kills it, and the one piece of the family the data has twice declined to refute), and this issue is the direction after it. Neither blocks the other, and #2881's own pre-registration already carries the negative prior this issue's evidence implies. Do not close #2881 as covered by this.

## What closed the previous line

#2879's Grid re-measure (`docs/experiments/gmm-cut/REMEASURE-2846.md`) found, while answering a different question, that the axis three studies have been working on is exhausted. Every variant paired against **the run's own production row** (fold-anchored κ=0.3), production arm, ramp 6–20, 267 cells:

| variant | Δ vs production | p |
|---|---|---|
| `pooled_sim_oracle` *(reads labels — the family's own bound)* | −0.0059 | **0.22** |
| `pooled_priorfree` *(#2836's winner)* | +0.0027 | 0.58 |
| `pooled_mid` | +0.0095 | 0.0012 |
| `pooled_gumbel_any_priorfree` | +0.0151 | 2e−6 |

The **label-reading oracle is no longer significantly better than what production already does without labels.** `pooled_sim_oracle` is the empirical rate-loss minimiser over the sim scores read with true labels, with no parametric form at all, so it bounds *every* rule that picks a threshold from that sim set — Gaussian crossing, Gumbel crossing, and #2881's tail quantile alike. Anchoring the mixture took the cut-rule axis with it.

Meanwhile the decomposition replicates to within 0.0007 of #2836 across four independent terms, and `transfer` — finite-sim-set estimation — is **+0.0407**, two thirds of the total +0.0623. That is where the cost actually is, exactly as #2836 predicted.

## The task

Design and run the study that attacks the **fit** instead of the cut. Concretely, at minimum:

1. **Characterise `transfer`.** It is the gap between the oracle cut of the *sim-set* mixture and the oracle cut of the *test* set — i.e. what the sim set fails to tell you about the pool, not what the rule fails to extract from the sim set. Establish what it scales with (sim-set size, prevalence, node count, drift across the trajectory) before proposing a remedy. Right now it is one large number with no structure attached.
2. **Pre-register against the base row, not a reconstruction.** `analyze_cut.py` now emits `agg/cut_contrasts_vs_base.csv` and `decisions.beats_production` / `ship_candidate` from it (#2884); use those. `pooled_mid` remains only as the historical #2836 contrast, and `beats_midpoint` no longer gates anything (#2846 is why). If the study adds a variant, add it to `SHIPPABLE` too — `decisions.unclassified_variants` will name it if you forget, because the allowlist gates the verdict.
3. **Do not trust the theory bench on this family.** It has now mis-sized the same family twice, at full power, with the survivorship bug fixed on both sides. Its generative model of region voting does not match VG scores. Any bench-only result is a hypothesis, not evidence.

## What is out of scope

The Gumbel **crossing**. #2846 is closed: `gumbel_any_*` stays as the measured variant, is not promoted, and gets no more spend. The surviving strand of the EVT work — the tail-level rule at α ≈ 0.16, median 0.158 with IQR ratio 2.38, inside the pre-registered stability bar of 3 — is **#2881's**, not this issue's; see the note at the top.

Refs #2846, refs #2836, refs #2879, refs #2881.

samggreenberg:
Closes #2883. Discharges #3187. Bears directly on #2884 and #3116.

Fresh worktree off `dev` `5ccd7597`. **552/552 cells COMPLETED, 0 failures, 0 zero-byte outputs**, 40 min on the `cpu` partition, **no GPU stage** — the pile supplied every embedding. Pre-registration (`docs/experiments/transfer-2883/PREREG.md`) was committed before the array was submitted. Report: `docs/experiments/transfer-2883/REPORT.md`.

## The assumption check came first, and three premises had moved

- **Confirmed.** The cut axis is dead: #2881, the one strand #2883 carves out, ran in #3130 and closed *negative* on both arms.
- **Corrected (#3187).** The issue quotes `transfer` = +0.0407 from a decomposition that dropped NaN-linked steps per column. Re-running today's fixed analyzer over #3130's cells gives **+0.0372 of +0.0557 = 67 %**, residual exactly 0, and **478/3921 steps (12 %) with no complete oracle chain** — a fact nobody had seen. Verdict unchanged, so #3187 is discharged.
- **Stale.** `launch_cut.sh` pins `CALIB_HEAD=linear`; `PRODUCTION_HEAD` is `linear_svm` since #3198. This run uses the shipped head.
- **Wrong as stated.** `pooled_sim_oracle` "bounds every rule that picks a threshold from that sim set" — it bounds *sim-set* loss, not *test* loss, which is what every table in this line reports.

## Verdict

| production arm, ramp 6–20 | |
|---|---:|
| `transfer` as the chain reports it | **+0.041 ± 0.002** |
| against a cross-fitted reference | **−0.001 ± 0.002** |
| the reference point's own optimism | **+0.041 ± 0.002** |
| real estimation error at the operating point | **0.0030** (7 %) |

`oracle_cost` is the minimum of the empirical cost **over the test sample it is then scored on** — `oracle_cut`'s own docstring calls it a lower bound, not a rule. Subtracting a sample minimum inflates every gap measured against it, and here it is the entire gap. `giraffe` seed 0 with 10 test positives: naive reference 0.020, cross-fitted 0.115.

All four pre-registered hypotheses land **on both arms independently**:

- **H1** `transfer` symmetry 0.070 in threshold units vs siblings 0.27–1.00 — a variance, not a bias.
- **H2** optimism share **1.02**; the term is gone against an honest reference.
- **H3** `a + b/n_pos` fits at R² 0.90, slope 0.39 ± 0.015. The axis is **positives**, decided across cells (within a cell the two candidate axes are proportional and fit identically — the selftest asserts fit quality does *not* separate them).
- **H4** both regularised estimators beat `pooled_sim_oracle` on held-out data (`smooth` −0.0038 ± 0.0011, p = 3e−5), so it is not a bound.

One pre-registered sub-prediction was wrong and is called out in the report: `bag` was predicted to lose and did not.

## What ships: nothing, deliberately

The label-free `bagfit_*` arm is a **clean negative on the production arm** (−0.0008, p = 0.22) and stays in `SWEEP_ONLY`, out of the ship gate, exactly as pre-registered — #2883 item 1 asks for the characterisation before the remedy.

## Also in this PR

- `preflight.sh`: the anchored-grid knob check now runs only when `CALIB_ANCHORED=1`, and **says so when skipped**. Checking it unconditionally made every non-anchored study declare a divergence it did not have, and a fictional declaration hides the real ones. Verified to still fire when the family is on.
- Two lessons: a rounded threshold is a display value, not an input (re-scoring at one moved the reference by 0.018 against a 0.037 term — caught by a sanity check before the array); and a login-node timing is a guess (36× slow, nearly cut this study's only shippable arm at design time).
- 19-check selftest with planted truth, incl. asserting the join-alignment check fires on a broken join.

## Recommended follow-ups

1. Give the decomposition an honest reference, or report the fourth term as a bracket. **The same defect is #3116's open complaint about `rule_inefficiency` / `calibration_shift` — one fix, not two.**
2. Caveat `decisions.family_headroom_exhausted`: it is set by a failure to reject at p = 0.10, from a premise this run refutes.
3. Generalisation check on `coco_val × dinov3_patch` — needs `REGION_VOTING_BY_DATASET["coco_val"]` flipped, which is #2905's open question, so deliberately not entangled here.

https://claude.ai/code/session_01M4b3aVes3Vggk6tdJiQoDo


samggreenberg:
**Full suite green on the GRID at the PR head** (`3ba8e355c`), run in its own worktree so it could not disturb the study's:

```
ALL 8924 TESTS PASSED (56 skipped, 2 xfailed, total: 8980)
  pyright:   OK
  pip-audit: OK
RUN PASSED (all gates green)
```

7:54 wall on `cpu`, `RUN_TESTS_EXIT=0`. Includes the two new suites:

- `tests_lib/detectors/test_transfer_rules.py` — 18 tests over the estimators, asserted against closed form or symmetry rather than golden numbers. Two are there because they are *surprising*: bagging the argmin **loses** where positives are starved (so a null on `bag` is not a null on the idea), and the cross-fitted oracle is more expensive than the sample minimum **for every seed**, since that is a fact about sample minima rather than about our data.
- `selftest_analyze_transfer.py` — 19 checks on fabricated cells with planted truth: the bracket recovers a planted optimism to 0.002, the learning curve recovers its planted intercept and slope, a planted cheaper-than-ERM estimator is found and a dearer one is not, the `bagfit_*` arms cannot be promoted, and the join-alignment check is asserted to **fire** on a deliberately misaligned frame rather than merely to exist.

The doc gate (`scripts/check-docs.py`) also passes with the new report and its five figures: `check-docs OK: 236 markdown files, 6 invariants, no drift.`

---

PR #3266, opened 2026-08-26
Slides: renumber, restructure and re-illustrate the calibration deck

### LINKED ISSUE #3265: Slides
More slides tweaks.

Presenter:
I want less meta-discussion of what the slides' goals or structure are. No "The names are bare on purpose." No "it opens the talk bare". Tell the presenter what they should present, not how you made the slide they're looking at.

This will help shorten your presenter notes, which is also necessary. Every slide's notes should fit on a single presenter page; if you're going over, shorten the notes, don't take a second page.

1:
If the title slide is unnumbered, let's think of it as slide0. Then the first outline is slide1. Let's consider every later reiteration of the outline as another animation of that same slide, so it's 1a, 1b, etc.

Presenter3:
No need to repeat the instructions immediately.

The outline sections should be shorter and more straight-forward. (These aren't punny titles like the slides.) 
VTSearch, Calibration, Preference, Regions, The End.
We're deleting the "What Is Still Open" section.

We can use the last outline frame as the end slide, instead of "Questions". Just show the outline with "The End" bolded.

A better outline title is " Outline & OutLyin' ". Note the ampersand and apostrophe.

4:
It sucks to talk about images without showing images. We should at least show some positive examples of the hard thing we're looking for. (Even better would be ALSO showing some hard negative examples that are close-but-not-quite what we're looking for.)
We recently had all that VT drama with "Book". We could show some easy books, but also some magazines and DVD cases and notebooks.

After establishing our example concept here, we should stick with the same one throughout. Replace all those "cat" examples with "book" examples.

10:
The solid red touching solid green is incoherent (especially to the color blind!). Instead, fill the red with the original red hatching we did, and fill the green with the original green hatching we did. (Don't draw the internal dividers of the histograms; treat the once-solid regions as now-hollow polygons.)

Is this the same red as before? It looks brown.

14: 
Same fill instructions as slide 10.

17:
I think this needs expansion. (We can always take more slide(s) without taking more TIME.)
Show an ordered row of images (not just checks/xs). This is what comes from the model. Continue the "book" example.
Show labels under those images. This is what comes from labels.
Show that there are lots of reasonable places to threshold this line, each with their own count of false positives & false negatives.
THEN we introduce the notion of a Cost Function, and that different functions minimize at different points.
Now introduce our "Inclusion" idea and what we expect from it, mapping "I" to cost function.

18:
The red/green block doesn't seem useful. As soon as we show that the k=-10 & k=10 line minimize at the same spot, we're done.

23:
Title should be "Prior Convictions".

27:
This isn't ready for a talk yet. All these ideas are in flux. Delete this slide.

30, 31, 32, 33, 34, 35:
Delete these.

samggreenberg:
Closes #3265

## Presenter notes

Rewritten throughout to say what to present rather than how the slide was made — no more "the names are bare on purpose", no more "it opens the talk bare". Every slide's notes now fit **one** presenter page: `build.py`'s own estimator says so for all 23 fragments, where nine of them used to spill onto a second or third.

## Numbering (item 1)

Numbering moves out of Marpit and into `build.py`: the deck sets `paginate: false` and the script draws the number, because the two rules the deck wants are rules about *fragments*, which Marpit cannot see.

- **The title slide takes no number and does not consume one**, so the first outline is `1`, not `2`.
- **A fragment shown several times is one slide shown several ways.** The outline is `1a … 1f` however far apart its six pages fall — Marpit can only hold the *previous* page's number, which is no help when the pages are eighty slides apart.

The audience deck now reads `1a, 1b, 2, 3, 4a…4i, 1c, 5a…` and ends on `1f`.

The speaker build follows: a showing gets only the notes naming *its* letters, plus (on the first showing only) any note naming none. That is item **Presenter3** — the outline's instructions no longer repeat immediately, and each showing carries the one line for the section it is entering.

## The deck (items 3, 27, 30–35)

- Sections renamed to plain names: **VTSearch, Calibration, Preference, Regions, The End**. Title is `Outline & OutLyin'`.
- "What Is Still Open" deleted, and with it the roadmap slide (item 27).
- The last outline frame, with **The End** marked, *is* the closing slide — the Questions slide is gone.
- Region voting is promoted out of backup to open the Regions section. Every other backup slide (30–35) is deleted.

## Book, everywhere (item 4)

Slide 2 stops talking about images without showing any. It now carries a figure of three books over three things that are books, or are not, or it depends — a magazine rack, a boxed set of DVDs, a spiral notebook. The concept runs through the rest of the deck: the ranking on the new slide 13 is the same photographs, and **both UI screenshots were re-shot as a real session hunting books.**

Caltech-101 has no book category, so `figs/src/coco_fixture.py` downloads COCO val2017 on demand and files it by subject, weighted toward the rectangular, printed and shelved (laptop, monitor, keyboard) so the near-misses are the ones the deck argues about. Nothing of the corpus is committed; the rendered figures and the two `.webp` frames are.

Which frames get voted Good is named rather than counted, because box area is not a proxy for "this photograph is about a book": COCO's largest `book` box is a game manual in a Wii case and its second is a cat on a bed — both frames the deck puts up as hard negatives two slides earlier.

## Figures (items 10, 14, 17, 18)

- **10 and 14** — the E-step panels swap solid rust meeting solid green for two **hollow hatched** regions in the deck's own Bad/Good hatching, drawn as polygons over the whole row so no bin dividers appear. The hue is unchanged and is the same `#b45309` the theme calls `.neg` and every other figure uses; it is amber-leaning on purpose (the pair `#b45309` / `#0d8a5f` survives deuteranopia in a way red/green does not). What made it read as brown here was the *area* — a large flat fill of it. Hatched, it reads as rust again. Say the word if you want the hue itself moved; that is a deck-wide change and I did not make it unasked.
- **17** — a new slide before *Pick Your Poison*: ten book/not-book photographs in the detector's order, then what they actually are, then three cuts anyone could defend with their false-positive and false-negative counts. *Pick Your Poison* then introduces the cost function and Inclusion as the answer to the question that slide refuses.
- **18** — the gauge row is gone from `calib-knob-flow`. The two cost curves lying on top of each other across the band already say the two ends of the knob minimise in the same place; three gauges said it again in a picture whose content is that the three pictures are the same one.
- **23** — retitled **Prior Convictions**.

## Docs

`slides/README.md` grows a *Numbering* section; `slides/STYLE.md` records that the outline's showings are one slide and that its last one closes the deck.

## Testing

`VTSEARCH_ALLOW_DELETIONS=1 ./run-tests.sh slides` passes (the env var covers the nine deliberate fragment deletions). The change is confined to `slides/`, which is the one case where that group is the complete gate. Both decks render; PDFs attached in the conversation.

https://claude.ai/code/session_016njwrdZPYbGotxBuA1XGgn

---
_Generated by [Claude Code](https://claude.ai/code/session_016njwrdZPYbGotxBuA1XGgn)_


samggreenberg:
Correcting my earlier note on the colour: the deck's negative hue is now **red**, `#b91c1c`, and the constant, the theme comment, the STYLE table and every presenter note call it red. "Rust" was a euphemism for a hue that was not red — `#b45309` is amber, which is why a large flat fill of it read as brown.

I defended keeping the amber on colour-blindness grounds in the PR body. That does not survive a check. Simulated against deuteranopia, `#b45309` and `#b91c1c` land on the same olive beside the green's slate — the amber bought nothing an actual red does not. What separates the two sides for a deuteranope is lightness and the hatch direction, and the E-step panels now carry both.

All 90 committed figures were regenerated; the deck re-renders clean.

---

Issue #3156, opened 2026-08-17
Vetted Genome


samggreenberg:
Looking through docs/reports/2026-08-17-overview-bench.html, there are some clear incorrect annotations in visual genome. In the "No buses, and no argument that the labels are wrong either", it shows image 498326.jpg and 3078.jpg, both of which have a clear bus, front and center. (286074.jpg has a couple buses too, but at least those are smaller.) We can't trust reports when their groundtruth isn't true.

So what should we do about this? I assume we should just start annotating our own corrections. (There are probably errors in BOTH directions? Not just missing buses, but also cars incorrectly labeled as buses, or whatever.) How should we record these, so that future experiments can best use our corrections?

And if we're going to do all this manual work to correct, we certainly don't want to annotate EVERY class; we'll just do the ones we'll want saved for our experiments. And we better make darn sure we're saving the classes that are most useful. What ARE the current classes used in vg_box__{s,m,l}? We want classes where each classes be well represented in all three vg_box_{size} datasets, where possible. (E.g. it'd be a more cohesive story if we look for background-buses, middleground-buses, and foreground-buses. How else should we choose categories?)

Luckily we have a great tool for human image annotation called VTSearch. :) But I don't want to hand-label every saved category over every saved image. I guess we build a e.g. bus detector and then label the ones sorta close to the boundary? We need to set up the infrastructure to run those annotations and save their results.

samggreenberg:
Partially addressed in #3157 — the prep pass (per-band supply scan + class shortlist), plus the analysis in [`docs/plans/vg-scale-bands-and-corrections.md`](https://github.com/samggreenberg/VTSearch/blob/dev/docs/plans/vg-scale-bands-and-corrections.md).

**Answering "what ARE the current classes in `vg_box_{s,m,l}`?"** — from the overview-bench re-run, 10 per band:

- `vg_box_small`: `nose` `glasses` `watch` `camera` `tip` `outlet` `drain` `mask` `mustache` `tusks`
- `vg_box_medium`: `hair` `shorts` `clock` `lamp` `truck` `backpack` `basket` `frisbee` `holder` `chairs`
- `vg_box_large`: `fence` `hill` `lady` `couch` `court` `walkway` `runway` `station` `intersection` `barn`

Zero overlap, and that's structural rather than accidental: a category is banded by its **median** voted-box area across all of VG, so it lands in exactly one band. The published small-vs-large gap therefore confounds box size with class identity.

[...]

PR #3269, opened 2026-08-26
Seed the autopilot from a text sort, disclaim the 28 reports that did not, and report #3156

samggreenberg:
Started as a report on the #3156 `vg_scale` overview grid. Writing it up turned over two harness defects, the second of which invalidates part of what every calibration-family study has measured — so the PR grew. Three commits, reviewable in order.

## 1. Autopilot seeding: crop → text sort *(the load-bearing change)*

The app's Autopilot seeds its first three Good votes from a **text sort** — the user types a query and votes down that ranking. `simulate_voting_iterations` takes that ranking as `seed_scores`, and `al_strategies`, `EVAL.md` and `voting_iterations` all describe the argument as *"similarity to the typed query"*.

`calibration/run_cells.py` and `max_patch/run_cells.py` passed **cosine to a crop of one boxed positive** instead, picked by `seed % len(candidates)`. No user produces that ranking. `vtscore/eval/seed_scores.py` — a builder for the real thing — had existed the whole time with exactly one caller, in a different study family.

Seeding is now text where the embedder has a text tower, and the app's **other** documented start — three random known-good examples — where it doesn't. DINOv3 has no `embed_text`, so it takes the second path; that is the app's real behaviour, not a fallback hack.

**Measured on `boat@medium` × `siglip` after the change:** first positive at **rank 1**, five in the top 20. The exemplar that crop-seeding made look like the dataset's hardest positive ranks **4006 of 7749** under the query a user would type — it would never have anchored a run.

Queries live in `calibration/experiment_config.EXPERIMENT_QUERIES`, **not** `EVAL_DATASETS`: `vg_scale` is an experiment fixture, not an app demo, and `test_all_eval_datasets_reference_demo_datasets` rightly enforces that distinction (it caught my first attempt). 12 texts over 36 banded cells — the band belongs to the cell, so someone hunting a small boat still types "boat".

Rows now carry `seed_mode` (`text` / `known_good`) and `seed_query` in place of `exemplar_id`. **The root cause was not that the seeding was wrong but that it was unnameable afterwards** — the study's own rows could not say how it had started.

## 2. A scoped caveat on 28 reports

Every study that ran through either `run_cells` gets a banner saying what to distrust (anything about *how a run starts* — starvation, stuck runs, `n_good`, early-trajectory cost) and what still holds (within-study contrasts, where every arm seeded identically — a shared baseline shift, not an arm-dependent one).

Affected families were determined from the **pre-fix** committed tree, since only two files ever carried `_load_exemplar`: `calibration`, `max_patch`, `fastproc`, `gpu_node`, `precision`. `inclusion_knob` and `mlp_vs_svm` are text-seeded and are left alone.

## 3. The #3156 report, and two preflight blind spots

`docs/experiments/vg-scale/REPORT_OVERVIEW.md` with figures. The grid ran 6480/6480 cells clean and measured the **retired `linear` head** — it launched from a worktree 321 commits behind dev, so `head=default (production)` resolved to the *previous* production.

**Why check 12 could not catch that:** it compares study pins against `PRODUCTION_*` constants read from the same worktree, so a stale checkout has a stale pin *and* a stale constant, they agree, and it passes honestly. A consistency check between two things that move together cannot detect them moving together. **Check 4 now fails a worktree >100 commits behind `origin/dev`.**

**Check 3** now counts header-only cells beside zero-byte ones. Its filter had only ever excluded `__sweep` though two more side frames had landed, so the first honest count read **12983 instead of 23**; now `! -name '*__*'`, and `-size -2k` narrows before the per-file `wc -l` (5.8s, not 88s).

**`launch_horizon.sh`** counts cells with `find`, not `ls`: 4 files per cell × 6480 blows `ARG_MAX`, `2>/dev/null` hides it, and `grep -vc` renders the failure as `0`. A finished grid reported `cells: 0 / 6480`.

Report findings, for what they are worth as a baseline arm: region voting wins cost by 0.10–0.12 and wins on the *ranking*; `calibration_shift` beats `rule_inefficiency` 8:1 (transfer, so no cut-rule work indicated); no cell is hard *as a cell*. The two "stuck exemplar" adjudications ended up being seeding artefacts — see commit 2's writeup — though `knife@small` exemplar **2322075** is a genuine mislabel worth cleaning (empty pavement, no knife in the frame).

## Testing

192/192 targeted tests (`al_strategies`, `autopilot_flow`, `eval`, `al_benchmark`, `docs_gate`), `ruff` clean, docs gate clean across 246 markdown files. Detectors validated against ground truth on the finished grid (23 header-only, matching `analyze_overview.py`'s independent count) and the new preflight check against a 354-behind worktree.

**No run was launched.** Job 549465 was cancelled at the owner's direction to free the overnight fleet; its 4813 completed cells are preserved, though being crop-seeded that grid wants a re-run rather than a resume.

## Not done

`coco_val` and `vg_box_{small,medium,large}` have no queries, so they now take the known-good start even on SigLIP where text is available. Config-only; the mechanism picks it up with no code change.

https://claude.ai/code/session_012Grv9aCvxhmYzgoxMPEhRZ


samggreenberg:
**Correction — both exemplars were adjudicated by eye, and they are not the same thing.**

Pushed in f9060d658, with both crops added as evidence figures.

- **`boat@medium` 2321462 — the label is CORRECT.** There *is* a boat: a hull on a cradle behind the tennis court fence, heavily occluded by a tree. At 1.16% of the image and in a scene whose every other cue says "tennis", it's a legitimately hard positive. Those 18 stuck runs failed honestly.
- **`knife@small` 2322075 — the label is WRONG.** The box lands on empty walkway beside a railway track; no knife in the box or anywhere in the frame. The same image also carries a spurious VG `pizza` box on the same pavement.

**The rule that built the shortlist doesn't hold.** I generated it from *"a model failure should not survive a change of representation"* — 2321462 refutes that. A positive that is tiny, occluded and out of context fails on every representation too, for the ordinary reason that it's hard. Both cases look **identical in every column this grid records**; only looking separates them. Cross-embedder persistence is a shortlist generator, not a verdict.

Revised accounting of the 266 stuck runs: **13 trace to the confirmed mislabel** and are recoverable by cleaning; **18 trace to the confirmed-real hard positive** and are not — that's the dataset working as intended. The other five repeat offenders are **unadjudicated** (all single-embedder, none looked at).

Recommendations reordered accordingly: drop/relabel 2322075, leave 2321462 alone, and adjudicate the remaining five by eye.

samggreenberg:
**Scope has grown substantially — this PR now also fixes how the autopilot is seeded, and disclaims 28 reports.** Pushed in `800e743cb`.

### The harness did not start runs the way the app does

Autopilot's Good phase seeds from a **text sort** — the user types a query and votes down that ranking until it has 3 positives. `simulate_voting_iterations` takes that ranking as `seed_scores`, and `al_strategies`, `EVAL.md` and `voting_iterations` all describe the argument as *"similarity to the typed query"*.

`calibration/run_cells.py` and `max_patch/run_cells.py` passed **cosine to a crop of one boxed positive** instead, picked by `seed % len(candidates)`. No user produces that ranking. `vtscore/eval/seed_scores.py` — a builder for the real thing — had existed the whole time with exactly one caller, in a different study family.

Seeding is now text where the embedder has a text tower, and the app's **other** documented start — three random known-good examples — where it doesn't. DINOv3 has no `embed_text`, so it takes the second path; that's the app's real behaviour, not a fallback hack.

**Measured on `boat@medium` × `siglip` after the change:** first positive at **rank 1**, five in the top 20. The exemplar that crop-seeding made look like the dataset's hardest positive ranks **4006 of 7749** under the query a user would type — it would never have anchored a run. Both of the "stuck exemplar" findings earlier in this PR were artefacts of the seeding, not properties of the images.

### Where the queries live

In `calibration/experiment_config.EXPERIMENT_QUERIES`, **not** `EVAL_DATASETS`. My first attempt put them in the latter and `test_all_eval_datasets_reference_demo_datasets` correctly rejected it: `vg_scale` is an experiment fixture, not an app demo dataset. 12 texts over 36 banded cells — the band belongs to the cell, so someone hunting a small boat still types "boat".

### Rows now say how they were seeded

`seed_mode` (`text` / `known_good`) and `seed_query` replace `exemplar_id`. The root cause wasn't that the seeding was wrong — it's that it was **unnameable after the fact**. The study's own rows couldn't say how it had started.

### 28 reports stamped

Every study that ran through either `run_cells` gets a scoped banner saying what to distrust (anything about how a run *starts* — starvation, stuck runs, `n_good`, early-trajectory cost) and what still holds (within-study contrasts, where every arm seeded identically). `inclusion_knob` and `mlp_vs_svm` are text-seeded and were left alone.

### Testing

192/192 targeted tests pass (`al_strategies`, `autopilot_flow`, `eval`, `al_benchmark`, `docs_gate`), `ruff` clean, docs gate clean across 246 markdown files. No run was launched — 549465 was cancelled at the user's direction with its 4813 completed cells preserved for a later resume.

### Still open

`coco_val` and `vg_box_{small,medium,large}` have no queries, so they now take the known-good start even on SigLIP where text is available. Config-only to fix; the mechanism picks it up with no code change.

Issue #3290, opened 2026-08-28
Ship a per-embedder calibration_fraction (0.3 single-vector / 0.5 patch), and update the slides


samggreenberg:
Following the #3287 measurement (PR #3289): make `calibration_fraction` a
**per-embedder default** — `0.3` when the detector learns in a single-vector
space, `0.5` when it learns on a patch grid — and bring the SCALE'26 slides up to
date with what the calibration work now says.

Recall the direction, because it inverts the reading if taken the other way:
`calibration_fraction` is the **Calibrate** share, so `0.3` means **70% Train /
30% Calibrate** — more of the user's votes go to fitting each fold's model, fewer
are held back to read its threshold from.

> **Note on the gate.** This issue originally asked for a *per-voting-mode* rule
> (0.3 whole-image / 0.5 region), mirroring `PRODUCTION_SCHEDULE_BY_MODE`. The
> data refutes that gate — see "Why not the voting mode" below. The requested
> behaviour is unchanged for every configuration that was measured to want it;
> only the predicate changes, and it changes to avoid a measured 3σ regression.

---

### The evidence

From #3287 (480 + 240 cells, `vg_scale_any`, 12 classes at identical prevalence,
4 seeds, 150 clicks; paired within `(dataset, category, seed, geometry)`,
bootstrapped over cells). Cost, paired vs the incumbent 0.5:

| geometry | learns in | voting | 0.3 | 0.4 | 0.6 | 0.7 |
|---|---|---|---|---|---|---|
| `siglip/whole_image` | single vector | binary | **−0.012 ± 0.003** | **−0.013 ± 0.003** | +0.003 ± 0.003 | +0.008 ± 0.003 |
| `siglip2_l/whole_image` | single vector | binary | **−0.013 ± 0.003** | −0.005 ± 0.003 | +0.006 ± 0.003 | +0.006 ± 0.003 |
| `dinov3_patch/whole_image` | patch grid | binary | **+0.015 ± 0.005** | −0.001 ± 0.004 | +0.005 ± 0.005 | +0.020 ± 0.005 |
| `dinov3_patch/max_patch` | patch grid | region | −0.002 ± 0.002 | −0.001 ± 0.002 | −0.001 ± 0.003 | +0.008 ± 0.003 |

`regret_honest`, against a cross-fitted reference, agrees on every row. For both
single-vector embedders 0.3 wins **in every vote band** — there is no regime where
[...]

Issue #3287, opened 2026-08-27
Sweep calibration_fraction per voting mode — is a 50/50 Train/Calibrate split actually optimal?


samggreenberg:
**Difficulty: medium.** Needs a GRID sweep; the code change is a one-line arm, the work is the study design and the reading.

**Recommended Claude model: Opus 4.8.** The implementation is trivial; the risk is in the analysis. This knob moves two things in opposite directions at once (fold-model quality vs. anchor/quantile resolution), which is exactly the anti-correlated-terms trap `_operating_metrics`' docstring already warns about for `calibrate_count` — a plausible-but-wrong reading here would ship a worse default.

## The question

`calibration_fraction` splits each calibration fold's labelset into a Train half (fits the fold model) and a Calibrate half (supplies the held-out scores the threshold is read from). It has been `0.5` since it was introduced, and as far as I can tell that value was never measured — it is the obvious default, not a result.

It is a genuine trade-off, not a "more is better" knob:

- **More Train** → better fold models → their haystack scores and held-out orderings are a better proxy for the final model's, but there are fewer anchors and a coarser conformal quantile.
- **More Calibrate** → more anchors for `fit_fold_anchored_cut`, finer conformal quantiles — but the fold models drift further from the final model (which always trains on *all* votes), so the scale the cut is read on is less like the scale it is applied on.

## Why it's worth measuring now

Investigating #3286 (the period-4 waves in learning curves) turned up **suggestive evidence that the optimum differs by voting mode.** The wave there comes from ±1-vote jitter in this very split, and the two geometries wobble in *anti-phase*:

- The whole-image arms (`siglip/whole_image`, `siglip2_l/whole_image`) are worst at the parity where the extra vote goes to **Train** — i.e. they are hurt by losing calibration data.
- The `max_patch` arm is worst at the parity where the extra vote goes to **Calibrate** — i.e. it is hurt by losing training data.

That is a one-vote perturbation, so it is weak evidence and it is confounded with the pick-order feedback loop. But it points the same way twice, and if it holds, the shipped constant is wrong for at least one of the two modes. Note the precedent: #2841 measured the blend schedule and found the two modes wanted *different* curves (`PRODUCTION_SCHEDULE_BY_MODE`), so a per-mode split fraction would not be a novel shape for this codebase.

## Proposed study

[...]

Issue #3278, opened 2026-08-27
Sixteen other launchers still open their DINOv3 arm on known-goods


samggreenberg:
#3276 fixed this for the `vg_scale` overview by giving its region arm a paired
embedder (`siglip+dinov3_patch`) and wiring `--require-text-seed` into
`launch_scale.sh`. **Sixteen other launchers carry a DINOv3 arm and none of them
passes that gate:**

```
launch_acq_region_2905.sh   launch_all.sh          launch_anchored.sh
launch_bench.sh             launch_cut.sh          launch_errdump.sh
launch_folds_2897.sh        launch_folds_3115.sh   launch_horizon.sh
launch_incl_2865.sh         launch_mixin.sh        launch_rate_2861.sh
launch_safe.sh              launch_safe_ab.sh      launch_tail_2881.sh
launch_transfer_2883.sh
```

This is **not** a claim about their existing results. Those all ran before PR
#3269, so every arm seeded from a crop — a shared baseline shift, already
disclaimed on the 28 stamped reports, and their within-study contrasts stand.

It is a claim about the **next** run of any of them. Post-#3269 a DINOv3 arm
opens on three random known-goods while every SigLIP arm opens on a typed query,
and where the study compares the two that difference is arm-dependent and does
not cancel.

## What each launcher needs is a judgement, not a sweep

The right answer differs, which is why #3276 did not change all sixteen:

* **Studies that contrast voting modes** (`launch_bench.sh`,
  `launch_folds_3115.sh`, `launch_anchored.sh`, …) want the pair, for the same
  reason #3156 did — otherwise the mode axis carries a seeding axis.
* **Single-arm region studies** (`launch_acq_region_2905.sh`) have no contrast
  to confound, but still would not open the way the app does. Whether that
  matters depends on whether the study's claim is about the opening.
* Some may legitimately want the known-good start, if that is the flow under
  test. Then they should say so, and pin it, rather than inherit it.

## Suggested shape

1. Add `--require-text-seed` to each launcher whose cells should be text-seeded,
   and the pair to its embedder list.
2. For any launcher that deliberately keeps the known-good start, record that as
   a pinned choice with a one-line reason, so the next reader does not have to
   re-derive whether it was intended.
3. Consider whether `region_voting_for`-style per-cell premise assertion should
   extend to the opening — i.e. a cell records what it *got*, and the analyzer
   refuses to pool two openings. `seed_embedder` (added in #3276) is the column
   that makes this checkable.

[...]

Issue #3115, opened 2026-08-12
Cross-calibration pools fold scores where the anchored path averages them — which is right?


samggreenberg:
**Recommended Claude model: Opus 4.8.** Adding the arms is mechanical, but choosing the combine semantics and reading a result whose terms are entangled is where this goes wrong quietly.

Refs #2897

## Two functions in this repo disagree about the same empirical fact

`threshold_from_fold_orderings` (the cross-calibration path the live app calls at `vtscore/state/core.py:1367`) **pools** every fold's held-out scores into one bag and takes a single conformal quantile. Its docstring justifies this:

> *"All folds' scores live on the same sigmoid scale, so the pool is exchangeable enough for the quantile rule."*

`FoldAnchoredCut._combined_fold_quantile` (the anchored path) does the opposite — one cut per fold, each converted to **that fold's own** quantile, then `qmean`/`qmedian`:

```python
quantiles.append(float(np.searchsorted(src, cut, side="left")) / float(src.size))
...
return float(np.mean(quantiles))
```

and its docstring says it keeps each fold's haystack specifically *"to read a cut's quantile in the scale it was measured on"* — i.e. it is built on the premise that fold scores are **not** directly comparable.

One of these is wrong. Nobody has measured which.

## Why pooling might be the wrong default

- **Quantile-of-mixture ≠ mean-of-quantiles.** Pooling estimates the quantile of the *mixture* of K different half-trained models. That mixture is wider than any single model's score distribution, so its quantile sits further into the tail — and the widening grows with K. That is a systematic (bias) effect of the fold count, not a variance one.
- **The resolution argument may be counting duplicates.** The docstring defends pooling because per-fold quantiles on a handful of votes are coarse. But the folds resample a *fixed* labelled pool at `calibration_fraction=0.5`, so `n_cal_scores` growing 6 → 100 across K ∈ {1…16} (measured in #2897) is largely the same few labels replicated. The extra resolution is partly illusory; averaging K cuts also refines resolution without pretending the sample grew.
- **Contamination vs dilution.** A degenerate fold (holdout with no positives) injects its scores straight into a pooled quantile. Averaging gives it 1/K weight; `qmedian` gives it ~none. #2897 measured `degenerate_rate` in region voting *rising* monotonically with K in the deep window — 0.0 at K≤2 to 0.0031 at K=16 — which is the signature this predicts.

## What to run

[...]

PR #3293, opened 2026-08-28
#3292: CLIP wants 0.3 too — the split optimum is not the SigLIP family, the vector width, or the capacity

### LINKED ISSUE #3292: Add CLIP to the pile and re-run #3287: a third single-vector embedder, from a different family
Add CLIP image/text vectors to the pre-embedded pile, then re-run the #3287
split-fraction grid on them to get a **third** single-vector data point.

**CLIP is exploration-only.** It is not a production candidate and must not become
a selectable app embedder — see "Keep it out of production" below. Its whole job
here is to be a different model lineage that we can measure.

## Why CLIP specifically, and what it does (and does not) settle

#3290 proposes gating `calibration_fraction` on the embedder: **0.3** when the
detector learns in a single-vector space, **0.5** when it learns on a patch grid.
That gate rests on two single-vector embedders — and **`siglip` and `siglip2_l`
are the same family.** So what #3287 actually established is closer to *"the
SigLIP family wants 0.3"* than *"single-vector embedders want 0.3"*, and #3290's
constant is an induction across one lineage.

CLIP is the cheapest way to test that. It is single-vector and language-aligned
like SigLIP, but a genuinely different lineage — OpenAI's softmax/InfoNCE
contrastive objective rather than SigLIP's pairwise sigmoid loss, different data,
different training recipe.

**What this run settles:**

| CLIP wants | reading |
|---|---|
| **0.3** | "single-vector ⇒ 0.3" survives a change of family. #3290's gate is on much firmer ground and can ship as written. |
| **0.5** | The gate as written in #3290 is **wrong** — the effect is narrower than "single-vector", and #3290 should not ship until we know what it actually keys on. |

That second row is the point: this is not just another data point, it can
**falsify the constant #3290 wants to ship**. Worth running before #3290 lands.

**What it does *not* settle.** CLIP has a text tower, so it opens on its own text
sort and its opening and learning happen in the **same space** — exactly like
`siglip` and `siglip2_l`. It is therefore a third *space-matched* arm and leaves
#3287's space-match confound untouched. The discriminator for that axis is still
the **`siglip+siglip2_l` pair** named in #3290. The two follow-ups are
complementary; neither replaces the other.

---

## Part 1 — a CLIP embedder

There is no CLIP embedder in the tree today (`vtscore/media/` holds the
implementations; there is no `vtscore/embedders/`). So this is a new embedder,
not just a pile row.

- **Template:** `vtscore/media/image/embedder_siglip.py`, with the shared
  cross-modal pieces in `vtscore/media/image/_cross_modal_shared.py`. Register
  through `vtscore.media.register_embedder` (`vtscore/media/__init__.py:262`).
- **Suggested model: `openai/clip-vit-large-patch14`.** Its image embedding is
  **768-d, the same as `siglip`**, which takes output dimensionality off the table
  as an explanation — a difference then cannot be "CLIP's vectors are wider or
  narrower". Not gated on HF, so no token dance (unlike `dinov3_patch`). If a
  different checkpoint is chosen, say why in the report; the family contrast is
  the point and capacity is the obvious confound.
- **It needs a working text tower**, since the study's cells must open on a typed
  query (`CALIB_REQUIRE_OPENING=text`). `embed_text_query` must return a vector
  for it, or every cell falls back to the known-good start and the arm silently
  stops being comparable to the SigLIP ones (#3278).

### Keep it out of production

- Do **not** add it to the app's embedder picker or to any default.
- It must **not** gain a `default`-kind mirror in `scripts/check-eval-app-sync.py`
  — that file is for constants the app actually ships.
- Note that #3290's predicate (`is_patch_embedder(learn_embedder)`) would classify
  CLIP as single-vector and hand it 0.3. That is harmless precisely *because* CLIP
  is eval-only, but it is worth a comment where the table lives so nobody later
  reads the eval arm as a shipped configuration.

## Part 2 — pile cells

- Add to `scripts/experiments/pile/pile_config.py`:
  `"clip": {"patch": False, "batch": 128}` (batch alongside `siglip`'s; tune if it
  OOMs).
- **`pile_config.cells()` is a full cross product** of `DATASETS × EMBEDDERS`, so
  adding the key schedules CLIP against all 8 datasets. That is fine and probably
  worth having — ~61k images total, and at `siglip`'s measured rate (7747 medias
  in 92 s) it is roughly **12 minutes of GPU and ~200 MB**. To scope it tighter,
  `build_pile.py` takes `--datasets` / `--embedders`:

  ```bash
  python build_pile.py --embedders clip --datasets vg_scale_any
  python build_pile.py --verify --embedders clip
  ```

- `vg_scale_any` is the **only** cell the study below needs.
- Run `--verify` afterwards: it asserts geometry rather than trusting the arm
  table, and it is what catches a single-vector cell being mistaken for a
  region-capable one.
- The pile is treated as purgeable, so the rebuild path has to work from staged
  sources — the usual rule, just don't forget the new row.

## Part 3 — re-run the #3287 grid

Identical to the `siglip2_l` follow-up in PR #3289, one env var different. The
launcher already derives its own preflight contract and sizing from the embedder
list, so a single-vector grid needs no other changes:

```bash
cd /exp/$USER/projects/<worktree>/scripts/experiments/calibration
export CALFRAC_BASE=/expscratch/$USER/calfrac-3287-clip
export CALIB_VGSCALE_EMBEDDERS=clip
bash launch_calfrac_3287.sh prepare
bash launch_calfrac_3287.sh baseline      # the click-0 anchor
bash launch_calfrac_3287.sh arms          # 5 arrays + one cross-arm analyze
```

**Measured sizing**, from the `siglip2_l` run in PR #3289 (same shape, same
dataset, same 12 classes × 4 seeds × 150 clicks):

| | |
|---|---|
| grid | 5 fractions × 48 cells = **240 cells** |
| per cell | ~3.1 min, ~0.9 GB peak |
| concurrency | 120 (5 arms × `%24` × 4G) — **CPU-bound, not memory-bound**, so no 12G patch floor |
| wall clock | **~12 minutes**, launch to report |

Add the pile build and this is well under an hour end to end.

Report the result **into the #3287 report** as a third single-vector row rather
than as a separate study — the whole value is the comparison, and a separate
document would drift from it.

## Acceptance

- [ ] CLIP embedder registered, with a working text tower (a cell opens on a typed
      query, not the known-good fallback).
- [ ] Not selectable in the app; no `check-eval-app-sync.py` mirror.
- [ ] `vg_scale_any × clip` in the pile and passing `build_pile.py --verify`.
- [ ] 240/240 cells, and the per-geometry table in
      `docs/experiments/calibration-fraction-3287/REPORT.md` extended with the CLIP
      row.
- [ ] A stated verdict on #3290's gate: does "single-vector ⇒ 0.3" survive a
      change of family, or does #3290 need rethinking before it ships?

Refs #3287, #3290. Evidence and the launcher this reuses: PR #3289.

samggreenberg:
Closes #3292. Evidence for #3290; extends the #3287 report rather than forking a new one.

## The result

Two CLIP checkpoints on the #3287 grid. **Both want 0.3 over the shipped 0.5.**

| geometry | cost @ 0.3 | cost @ 0.4 | cost @ 0.6 | cost @ 0.7 |
|---|---|---|---|---|
| `clip/whole_image` (ViT-B/32, 512-d) | **−0.011 ± 0.003** | **−0.0087 ± 0.0031** | +0.0065 ± 0.0035 | +0.010 ± 0.0033 |
| `clip_l/whole_image` (ViT-L/14, 768-d) | **−0.016 ± 0.003** | **−0.014 ± 0.003** | +0.0058 ± 0.0034 | +0.021 ± 0.0041 |

Negative is better than 0.5, `regret_honest` agrees independently, and the sign structure is the same in all four vote bands of both arms: 0.3/0.4 below zero in eight of eight band-arms, 0.6/0.7 above it in eight of eight.

**480/480 cells, zero failures, zero zero-byte outputs, 7,068 production rows in every one of the ten arms — nothing dropped.** Every metric row opened on a typed query in the arm's own space, so neither arm fell back to the known-good opening that would have made it incomparable to the SigLIP rows (#3278).

## The verdict on #3290

The issue pre-registered the reading and the answer is the first row: **"single-vector ⇒ 0.3" survives a change of family, so #3290's gate can ship as written.**

It survives it *twice*, which is more than was asked for. **I ran two CLIP checkpoints, not one** — leaving SigLIP changes family, vector width and encoder capacity all at once, which is exactly the confound that turned #3115 into a wrong law. Two capacities of one lineage separate them:

- **not the family** — a different objective, data and recipe still wants 0.3;
- **not the width** — 512-d and 768-d agree, and the 768-d arm is the *stronger* of the two;
- **not the capacity** — ViT-B/32 and ViT-L/14 agree.

`clip_l` is `openai/clip-vit-large-patch14` precisely because its 768-d output matches `siglip`'s.

## What it does not settle, and this got *worse*

**The space-match confound is untouched.** Both CLIP arms open on their own text sort, so opening and learning happen in the same space — like `siglip` and `siglip2_l`, unlike both `dinov3_patch` arms. The split is now **four space-matched arms wanting 0.3 against two space-mismatched arms wanting 0.5**, and "space-match" still explains all six geometries exactly as well as "single-vector" does. Adding a fourth arm on the same side of that line cannot break it.

The discriminator is the **`siglip+siglip2_l` pair** — single-vector *and* space-mismatched, the one combination never run. ~20 minutes, and now the highest-value cell in the design.

So #3290's constant is right for every configuration measured, and the predicate it is written on (`is_patch_embedder`) is still not known to be the predicate doing the work.

## Two things worth a reviewer's attention

**The issue's Part 1 premise was wrong.** It says there is no CLIP embedder in the tree. `vtscore/media/image/embedder_clip.py` (`clip-vit-base-patch32`, 512-d) has shipped for a while and *is* app-selectable, so "do not add it to the picker" described a state that no longer held. I left that embedder exactly as it is — withdrawing a choice users already have is a production decision, not an experiment's to make — and used it as the second capacity instead.

**`eval_only`, and why it is a code property rather than a setting.** `clip_l` is a research arm and nothing has evaluated it *for the app*, so `MediaEmbedder` gains an `eval_only` property that `embedders_for_type` (every picker and per-media-type default) and `all_embedders_dict` (`GET /api/embedders`) withhold. The existing `hidden_plugins` mechanism would have worked but is a *deployment setting* someone has to apply; this cannot be reached by a deployment forgetting. Resolution by name stays open deliberately — a pile cell embedded by an eval arm still has to load.

## Also here

- `smoke_clip_3292.py` gates the pile build on the failure that would otherwise be silent: no text tower means cells open on the known-good fallback and *nothing raises* (#3278). It also pins each arm's dimension and checks the text/image spaces are actually aligned.
- `run_3292.sh` chains the pile build and both grids on `afterok`, so the run needs no terminal — every laptop-side watcher this project has run has eventually died with the VPN. It was in fact launched, disconnected from, and read back off disk.
- **A sizing correction, reported rather than buried.** This run asked for 8G per task instead of the launcher's 4G single-vector default, reasoning that `run_cells.py` embeds the text query per cell so CLIP ViT-L/14's weights would sit in every cell's RSS. Measured peak: **1.16 GB**. The default had 4× headroom and the halved concurrency bought nothing. The reasoning was sound and the number was not.

## Cost

Pile build 8m 24s; each study ~19 min prepare→report; **47 min end to end**. Per cell 3.1 min (1.8–3.9).


samggreenberg:
**Suite is green.** `9176 passed, 0 failed, 56 skipped` on the GRID (job 581118, `RUN_TESTS_EXIT=0`); pyright OK, pip-audit OK, `check-docs` OK across 256 markdown files, eval/app sync 13 mirrors current. Frontend build/unit tests were **skipped** — no `node_modules` in the worktree — which is worth knowing, though this branch touches no frontend code.

Four commits landed after the description above was written. Two are worth a reviewer's attention:

**`eval_only` is deliberately not on the wire.** The first cut put it in `to_dict()`, and the suite returned 21 failures — 18 of them `to_dict` exact-dict assertions across every embedder. I removed the key rather than editing eighteen tests: `all_embedders_dict()` filters eval-only embedders out, so it could only ever serialise as `False`. A constant in the API contract is not worth that. The property stays; ask the embedder, not its dict.

**The two registration tests that remained are the interesting ones.** `all_embedders` is the *registry*, so its count goes 25 → 26 and its name set gains `clip_l`. The `embedders_for_type("image")` set immediately below it is **unchanged** — and that test passing untouched is the actual evidence the withholding works, so the comment now points at it rather than restating the count.

The other two were gates catching me: an em-dash in a heading broke its own anchor (`check-docs` slugifies it away), and I hand-edited `LESSONS.md`, which is generated and sorts within a date.

---

Also in here, from an incident during the run rather than from the science: `scripts/experiments/lessons/2026-08-28-a-detached-worktree-ate-a-commit-and-the-push-said-ok.md`.

`/exp/$USER/suite.sbatch` runs `git checkout --detach` in the worktree you hand it and did not put it back. A fix committed afterwards landed on **no branch**; `git push <branch>` then exited 0 having pushed nothing, and the next suite run discarded the commit by checking the branch out again. It only surfaced because the suite failed the *identical* way twice.

Recovered from the reflog. The note "suite.sbatch leaves the worktree detached" already existed and did not help — it is a fact you have to recall three steps later, which is not a control. So `suite.sbatch` now restores the original ref in an `EXIT` trap (backup at `suite.sbatch.bak.20260828`; verified restoring on the last two runs). That file lives on `/exp` and is not in the repo, so the fix is **not** in this diff — only the lesson is.

The generalisable half, which cost the time: **`git push` exiting 0 does not mean your commit is on the remote.** It means the push had nothing to complain about, including nothing to do. `git rev-parse HEAD origin/<branch>` — two identical shas — is the assertion that distinguishes them.

Issue #3305, opened 2026-08-28
box_sheets.py looks for COCO pixels in a directory that has never existed, and draws an empty sheet instead of saying so


samggreenberg:
Found while fixing the `coco_val` false alarm in #3299. Same root cause, different consumer, and this one has no symptom at all.

## The problem

`scripts/experiments/pile/box_sheets.py` resolves COCO pixels through `pile_config.COCO_IMAGES`:

```python
IMAGE_DIRS = {
    "vg": [pc.DEMO_CACHE / "visual_genome" / "VG_100K", pc.DEMO_CACHE / "visual_genome" / "VG_100K_2"],
    "coco": [pc.COCO_IMAGES],          # /exp/scale26/datasets/external/COCO/images/val2017
}
```

That directory does not exist and never has. The staging area holds the zip:

```
$ ls /exp/scale26/datasets/external/COCO/images/
train2017.zip   val2017.zip
```

`_load_coco` knows this and reads `val2017.zip` directly (now `pc.COCO_VAL_ZIP`, after #3299). `box_sheets.py` does not.

## Why it is silent

`path_of` walks its candidate dirs and returns `None` when nothing resolves:

```python
for d in dirs:
    p = d / Path(str(cand)).name
    if p.exists():
        return p
```

Every COCO media takes that path, so the sheet is generated with no thumbnails rather than failing. Nobody has noticed because `--dataset` defaults to `vg_scale` and every sheet made so far has been a VG one.

This matters more than a cosmetic bug because of what contact sheets are *for*: the standing report guidance is that any claim about a **kind** of error owes literal examples, and on image data that means the picture — it is the only way a reader can separate a model error from an annotation error. A sheet that silently omits every image answers the question with nothing while looking like an answer.

## Suggested fix

Two parts, and the second is the one that matters:

1. Read COCO pixels from `pc.COCO_VAL_ZIP` the way `_load_coco` does (`zipfile.ZipFile`, member basename lookup), keeping `pc.COCO_IMAGES` as an optional extracted-directory fast path if anyone ever unpacks it.
2. **Fail loudly when no image source resolves.** A sheet with zero thumbnails should be an error, not a file. The general shape here is the same one #3299 recorded twice today: a check, or a rendering, that names a different path than the build reads is worse than none, because it produces confident output either way.

## Not urgent

[...]

Issue #3297, opened 2026-08-28
The vg_box_* rebuild path has been broken since 2026-08-17: the scan file predates its own envelope


samggreenberg:
The pile documents itself as purgeable — `pile_config.py`: *"every cell must be rebuildable from sources that are **not** on scratch."* That invariant is currently **false for `vg_box_small`, `vg_box_medium` and `vg_box_large`**. If scratch were purged today, those three datasets would not come back.

Found incidentally while backfilling the CLIP columns for #3292 (the study itself needs none of these cells).

## The failure

```
python build_pile.py --datasets vg_box_small --embedders <any>
...
  File "scripts/experiments/pile/build_pile.py", line 194, in _band_categories
    stats = json.loads(scan_path.read_text())["categories"]
KeyError: 'categories'
```

**It is embedder-independent.** `_load_vg_band` calls `_band_categories(band)` before it touches the embedder, so `vg_box_small × siglip --force` fails identically. Nothing about CLIP is involved; the CLIP backfill just happened to be the first thing to ask for a `vg_box_*` rebuild in months.

## The cause: a scan-format change that outran the file

`scan_vg_boxes.py` returns `{"meta": {…}, "categories": stats}`. `_band_categories` reads `[...]["categories"]`.

The file on scratch, `/expscratch/$USER/vts-cache/vg_box_scale.json`, is the **bare `stats` dict with no envelope** — 4628 categories at top level.

| | |
|---|---|
| `vg_box_scale.json` on disk | **2026-08-12**, pre-envelope |
| envelope introduced | **2026-08-17**, `fb4f4ec03` (*"vgscan: per-band supply, so a class can span all three size bands"*) |
| `vg_box_*` cells built | 2026-08-12, against the then-current code |

So the rebuild path has been broken since 2026-08-17 and nobody noticed, because the built cells kept loading fine. **A cell that loads says nothing about whether it can be rebuilt** — which is the general shape worth taking away here, and an argument for exercising the rebuild path on something other than a purge.

## Why the fix is cheaper than it looks

`fb4f4ec03` also added `bands`, `bands_compact`, `n_compact` and `compact_frac` per category, none of which the old file has. But `_band_categories` reads only:

- `voted_area`
- `n_images`
- `union_inflation`

and **all three are present in the 2026-08-12 file**. The envelope is the only hard incompatibility.

That matters because it separates two very different repairs:

[...]

PR #3306, opened 2026-08-28
Build the last six pile cells, and make the canary and the GPU picker true (#3299)

### LINKED ISSUE #3299: Build the six vg_box_* × CLIP cells, and prove the pile is rebuildable against the real scan
The `vg_box_*` rebuild path is fixed in code (#3297), but the fix was verified
*statically* — nothing has yet run against the actual pile on scratch. This is
the GRID half, split out because it needs machine time and cannot be done from a
Claude session.

**Recommended Claude model: Sonnet 5.** The commands are documented and the
judgement calls are narrow; the one place to be careful is reading the
reproduction check in Part 3 honestly rather than assuming it.

Blocked on #3297 merging to `dev`.

## Part 1 — the six cells that were blocked

`vg_box_{small,medium,large} × {clip, clip_l}` are the six cells `build_pile.py
--list` reports MISSING. They were unbuildable because band selection died with
`KeyError: 'categories'` before the embedder was ever touched; with #3297 in
they should build.

```bash
python build_pile.py --datasets vg_box_small,vg_box_medium,vg_box_large --embedders clip,clip_l
python build_pile.py --verify --embedders clip,clip_l
```

These are single-vector embedders, so the cells are not region-capable and
`--verify` should say so rather than flagging a patch gap.

Not blocking any study today — #3292 needed only `vg_scale_any × clip`. This is
filling in the cross product so the columns are complete.

## Part 2 — run the canary against the real pile

#3297 adds `build_pile.py --rebuildable`: it runs each dataset's selection step
against its real sources and embeds nothing, so it costs seconds. It has only
ever run against synthetic scans in the test suite. Run it once on the GRID:

```bash
python build_pile.py --rebuildable
```

Expect all eight datasets OK. Anything that reports `REBUILD-BROKEN` is a second
purge-time landmine of the same kind as #3297 and wants its own issue — the
whole reason the mode reports every dataset instead of stopping at the first.

## Part 3 — confirm the repair reproduces the published sets, on the real file

#3297's third checkbox asked for this rather than an assumption. It was answered
statically and, as far as a laptop can go, conclusively: every input to selection
is byte-identical to what built the cells on 2026-08-12 — `BOX_BANDS`,
`BAND_MIN_IMAGES`, `BAND_MAX_INFLATION`, `BAND_N_CATEGORIES`,
`is_object_category`, `NON_OBJECT_CATEGORIES`, and the stratified-slot algorithm
itself. The only change to the whole path is the `["categories"]` subscript.
`docs/experiments/gpu-node-3160/REPORT.md` §6.1 corroborates it empirically: fed
the archived scan, today's selector reproduced all 40 categories per band and the
rebuilt cells carried identical media ids.

What is left is checking it against the file actually on scratch, which no
session has been able to read:

```bash
python -c "import build_pile, json; print(json.dumps({b: build_pile._band_categories(b) for b in ('small','medium','large')}, indent=1))"
```

Compare each list against the categories the live `vg_box_<band>__siglip.pkl`
holds. All three should match exactly. **If any band differs, stop and say so** —
that would make this a dataset change rather than a bugfix, and the numbers
published in #3129 and #3156 would need re-examining.

## Part 4 — a periodic canary

#3297 makes `--rebuildable` cheap enough to run constantly; nothing yet runs it
constantly. Consider hanging it off whatever already touches the pile on a
schedule, or off the front of `launch_pile.sh`, so a rebuild path cannot rot for
eleven days again. A purge is the worst possible time to discover it has.

## Acceptance

- [ ] Six `vg_box_* × {clip, clip_l}` cells built and passing `--verify`.
- [ ] `--rebuildable` green for all eight datasets against the real sources.
- [ ] The three band category lists confirmed identical to what the live
      `vg_box_*` cells hold — or the difference reported, loudly.
- [ ] A decision on where the periodic canary runs (or an explicit "not worth it").

Refs #3297.

samggreenberg:
Closes #3299.

The pile is complete: **40 of 40 cells**, `--verify` green on all of them. The six `vg_box_* × {clip, clip_l}` cells built on jobs `582384`–`582386`, 8m09s each, 0 failures — `clip` 136s embed per 12000 images, `clip_l` 229s.

The cells were the cheap part. Three things had to be repaired to get them, and each is worth more than the cells are.

## The GPU picker was reporting every GPU on the cluster as free

`pick_gpu.py` measured usage from `GresUsed=`. This cluster writes that field on **zero** nodes — `scontrol show node --oneliner | grep -c GresUsed=` returns `0` — and the reader flattened the missing field to an empty string, so `free == total` on every node, on every type, always. Rule 2 of its selection order ("fastest type with `--need` free") therefore always matched the **first** candidate. Since 2026-08-17 it has been a hardcoded `a100` wearing a query.

It looked like it was working:

```
pick_gpu:   a100      23 free /  23 total
pick_gpu:   l40s      14 free /  14 total
pick_gpu:   v100     110 free / 110 total
pick_gpu: a100 -- 23/23 free -- fastest type with 3 free to start on now
```

Every number there is the node total. Read off `AllocTRES` at the same moment: `a100 0/23`, `l40s 2/14`, `v100 109/110`. The three build jobs went to A100 and came back `Reason=Priority`, `StartTime=2026-08-29T12:33` — **a 24-hour wait, with 109 V100s idle**. Cancelled, resubmitted on V100, running in 20 seconds.

`_node_gpus_used` now prefers `GresUsed` where a cluster writes it, falls back to `AllocTRES`'s `gres/gpu:<type>=<n>`, and returns `None` when **neither** is present — an unmeasurable node contributes nothing at all, exactly like a drained one, so an unreadable cluster degrades to the fallback instead of confidently claiming everything is free. A missing measurement must not be spellable as a measurement of zero; both are falsy, and the failure is silent in the direction that reads like good news.

Verified on the cluster after the fix, picking `l40s` over `a100` for the first time ever:

```
pick_gpu:   a100       0 free /  23 total
pick_gpu:   l40s       1 free /  14 total
pick_gpu:   v100     110 free / 110 total
pick_gpu: l40s -- 1/14 free -- fastest type with 1 free to start on now
```

Twelve green parser tests missed this because **every fixture wrote a `GresUsed` field** — they described a cluster the code never meets. The new ones are copied verbatim off `rack5n06` and `rack10n01`.

## The canary's first real run was a false alarm

`--rebuildable` reported `coco_val` REBUILD-BROKEN against sources that were entirely intact. It checked `images/val2017`, an extracted directory the staging area has never held; `_load_coco` opens `images/val2017.zip`, 815 MB, present since July. The builder carried its own inline copy of that path while `COCO_IMAGES` named the directory, which is what let the two disagree.

`pile_config.COCO_VAL_ZIP` now names it once and both sides go through it. Three tests cover the COCO branch, which had none — including a source-level assertion that fails if the builder ever spells `val2017.zip` inline again, because the thing worth pinning is the *identity*, not the behaviour.

Worth stating plainly: **a false alarm costs exactly what a true alarm costs**, and it is spent on nothing. The issue instructed that any `REBUILD-BROKEN` "wants its own issue"; filing one would have been the reasonable-looking wrong move.

## The canary asked a weaker question than it looked like it asked

That selection *runs*, not that it selects the same thing. Those come apart in the direction that hurts: #3297 had two candidate repairs, both of which made the selector run again, and only one of which kept picking the categories the published `vg_box_*` sets hold. The other would have redefined three datasets with the right media count, the right vectors and nothing to look at that would say so.

So `--rebuildable` now compares today's selection against the vocabulary of the smallest built cell (`_band_vocab_drift`), and reports `REBUILD-BROKEN` on any difference. That is #3299's Part 3 answered by committed code rather than by hand — and against the live pile it passes:

| dataset | selector picks | live cell holds | cells compared | verdict |
|---|---:|---:|---|---|
| `vg_box_small` | 40 | 40 | `siglip`, `siglip2_l`, `dinov3_patch` | identical |
| `vg_box_medium` | 40 | 40 | `siglip`, `siglip2_l`, `dinov3_patch` | identical |
| `vg_box_large` | 40 | 40 | `siglip`, `siglip2_l`, `dinov3_patch` | identical |

Not just against `siglip`: all three present cells of each band agree with each other and with the selector, and every selected category is carried by at least 52 images in its cell. So the #3297 repair is confirmed to be a bugfix and not a dataset change, and #3129 / #3156 need no re-examination. Costs 0.93s for all eight datasets, up from 0.23s.

## Where the periodic canary runs

In front of every launch, inside `launch_pile.sh`'s existing blocking CPU stage. Every build already touches the pile, the check is sub-second, and a purge is the worst possible moment to learn the rebuild path rotted.

Two runs, deliberately: the first reports **all eight** datasets, so rot under something you are not building today still gets seen; the second covers only the datasets being submitted and its exit code gates the launch, because a broken source under a dataset nobody asked for is news rather than grounds to refuse. Both halves tested against the real launcher on the cluster:

- `VTS_COCO_ROOT=/tmp/nope bash launch_pile.sh coco_val` → `REBUILD-BROKEN`, `prelaunch FAILED -- nothing submitted`, exit 1, empty queue.
- `VTS_COCO_ROOT=/tmp/nope bash launch_pile.sh vg_box_small` → same report, launch proceeds, `submitted vg_box_small -> job 582411`, exit 0.

It runs inside the `srun` rather than on the login node because `build_pile.py` asserts it is running against its own checkout's `vtscore`, which needs the venv.

## Also here

- `README.md`'s grid section said "three datasets x three embedders" and listed six datasets. It is eight × five now, with a table, and a note on what the CLIP columns buy that the SigLIP pair cannot (family at fixed capacity, rather than generation and capacity moving together).
- Two lesson files, and `LESSONS.md` regenerated.
- #3305 filed: `box_sheets.py` is the other consumer of the non-existent `COCO_IMAGES` directory. It does not error — it draws an empty sheet. Not urgent (no COCO sheet has ever been asked for), not fixed here.

## Testing

`tests_lib/core/test_pick_gpu.py` 32 passed, `tests_lib/core/test_pile_box_scan.py` 17 passed. Full suite on the GRID: job `582413` — result added below.

### DIFF
--- a/scripts/experiments/LESSONS.md
+++ b/scripts/experiments/LESSONS.md
@@ -72,10 +72,12 @@ correct union of the two branches with no judgement involved.
 
 | Date | Lesson | Issue |
 |---|---|---|
+| 2026-08-28 | [a canary that checked a path the build never reads](lessons/2026-08-28-a-canary-that-checked-a-path-the-build-never-reads.md) | #3299 |
 | 2026-08-28 | [a committed figure is *output*, and git merged it clean](lessons/2026-08-28-a-committed-figure-is-output-not-a-file.md) | #3280 |
 | 2026-08-28 | [a detached worktree ate a commit, and `git push` said OK](lessons/2026-08-28-a-detached-worktree-ate-a-commit-and-the-push-said-ok.md) | #3292 |
 | 2026-08-28 | [a figure that hung, and three fixes that were not it](lessons/2026-08-28-a-figure-that-hung-and-three-fixes-that-were-not-it.md) | #3156 |
 | 2026-08-28 | [I nearly rebuilt a pile dataset another session was rebuilding (#3281, #3284)](lessons/2026-08-28-nearly-rebuilt-a-dataset-another-session-owned.md) | #3281, #3284 |
+| 2026-08-28 | [the GPU picker reported every GPU on the cluster as free](lessons/2026-08-28-the-gpu-picker-reported-every-gpu-free.md) | #3299 |
 | 2026-08-27 | [a box normalised twice](lessons/2026-08-27-a-box-normalised-twice.md) | #3281 |
 | 2026-08-27 | [a repair moves more than the defect](lessons/2026-08-27-a-repair-moves-more-than-the-defect.md) | #3281 |
 | 2026-08-27 | [The pick log was being concatenated into every analyzer's metric frame](lessons/2026-08-27-the-pick-log-was-in-every-metric-frame.md) | — |
--- a/scripts/experiments/lessons/2026-08-28-a-canary-that-checked-a-path-the-build-never-reads.md
+++ b/scripts/experiments/lessons/2026-08-28-a-canary-that-checked-a-path-the-build-never-reads.md
@@ -1,0 +1,50 @@
+# 2026-08-28 — a canary that checked a path the build never reads (#3299)
+
+**What broke.** `build_pile.py --rebuildable`, added the day before by #3300 to
+prove every pile dataset can still be rebuilt from off-scratch sources, ran for
+the first time against the real cluster and reported:
+
+```
+[pile]   coco_val           BROKEN   coco_val: missing /exp/scale26/datasets/external/COCO/images/val2017
+[pile] REBUILD-BROKEN: coco_val: ...
+```
+
+Nothing was broken. `_load_coco` reads
+`COCO_ROOT/images/**val2017.zip**` — 815 MB, present, staged in July and
+untouched. `val2017/` is an extracted directory the staging area has never
+held, and the README says so: *"COCO from the staged zip plus flattened
+annotations."* The canary checked `pc.COCO_IMAGES`; the builder spelled the zip
+path inline. Two names for one source, and the check took the one the build
+never opens.
+
+**Cost.** ~10 minutes of investigation, and it very nearly cost more: the issue
+that commissioned the run said anything reporting `REBUILD-BROKEN` "is a second
+purge-time landmine of the same kind as #3297 and wants its own issue". Filing
+that issue would have been the reasonable-looking wrong move. The check that
+settled it was two lines — read what `_load_coco` opens, then `ls` it.
+
+**The general form.** A canary is only as good as the identity between what it
+checks and what the real path does. Here the constant `COCO_IMAGES` existed and
+looked authoritative, so the new code used it without asking whether the builder
+did. **The builder had its own copy of the path**, which is what let the two
+disagree — and neither was wrong on its own terms.
+
+The consequence is worse than a plain bug. A false alarm costs exactly what a
+true alarm costs, and it is spent on nothing; a canary that cries wolf on its
+first real run is a canary people learn to skip, which is the one failure mode
+that makes it worthless. So a false positive here is not a milder version of a
+false negative. It is how you lose the check.
+
+**Prevented.** `pile_config.COCO_VAL_ZIP` now names the zip once; `_load_coco`
+and `rebuildable()` both go through it, and `COCO_IMAGES` is documented as the
+optional extracted directory that only `box_sheets.py` wants. Three tests in
+`tests_lib/core/test_pile_box_scan.py` cover the COCO branch of the canary,
+which had none: a staged-zip-only tree passes, a missing zip still fails, and a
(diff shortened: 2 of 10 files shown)


samggreenberg:
**Full suite green on the GRID** — job `582413`, `COMPLETED 0:0`, 8m32s.

```
9211 passed, 56 skipped, 1 xfailed, 1 xpassed in 424.50s
  pyright: OK
  pip-audit: OK
RUN PASSED (all gates green)
```

Issue #3325, opened 2026-08-30
Skylines and Baselines


samggreenberg:
I want to polish our standard Report Viewer and then rebuild it for recent experiment reports (especially the vg_scale survey).

We have this new Skyline point for how well an SVM *could* do on *all* the training data (using ?some threshold). We have the Baseline point, showing how the original text sort did with a GMM threshold. We also have the calibration oracle line, where we use the momentary model but with a cheating threshold. And of course we have the actual performance line, of how the momentary model did with its calculated threshold. (Note that some of those are POINTS and some of those are LINES.)

I'm interested in your take, but my idea is to put the Baseline point as a notch on the left and the Skyline point as a notch on the right. Abandon the current dotted line from the baseline point to the actual SVM performance; it's distracting and irrelevant, so leave that as a gap instead. And then for the oracle threshold line, I'd like a binary checkbox at the top for whether to draw the oracle as a dotted line along with the performance as a solid line.

I like the shadow that's currently displayed for the range of the population under the displayed average line. The problem is, it only works for one bold line on the graph. I'd like the ability to draw multiple bold lines on the same graph for comparison. Whether that's bikes vs birds or SigLIP vs CLIP, I want a Boolean checkbox that will take the multiple shadowed single-line graphs all in blue, and instead make one shadowless multiline-graph with a rainbow of hues to distinguish the lines from each other. (And a legend to understand what's what.)

First implement the above. Use the Question tool if you have any questions. Then, if you have enough data in the repo already, update the existing experimental reports where available. Then, if you need to do any light rerunning on the GRID, make a new Issue for that rerun GRID work.

samggreenberg:
Addressed in #3327.

The viewer rework is in and every committed `viewer.html` is re-skinned onto the new template. The GRID work you asked me to split off is #3326 — no committed study's payload carries an oracle cut or a skyline, and neither can be recovered from the page, so those two need the cell CSVs. The oracle half is a re-analysis only (`oracle_cost` / `oracle_fpr` / `oracle_fnr` / `n_test_pos` / `n_test_neg` are already on every base row ever emitted); only the skyline needs an actual re-run, and it is one extra fit per arm per cell.

---
_Generated by [Claude Code](https://claude.ai/code)_

samggreenberg:
Shipped to `main` in the 2026-08-31 release — fixed in #3327.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #3321, opened 2026-08-29
Training Regret


samggreenberg:
We pay a lot of attention to "Calibration Regret" when experimenting and exploring calibration. We "cheat", using the oracle's threshold of the test data using the model's current ranking. We could never attain this in practice, but it's sometimes useful to compare our mortal attempts vs the oracle threah. Should we also extend this idea to TRAINING regret? 

We could "cheat" by using the test data and to train an SVM, and then compare our mortal SVM vs that one. (We could also go halfway, and train an SVM on the ENTIRE train set, when generally we only get the 10-200 clicks from the simulated user.)

Would these tools be useful for comparing our trained models vs the possible trained models out there? What am I missing about how this idea would work? If it's useful to measure "learnability" (instead of "how much we've learned"), we can redo e.g. the vg_scale survey to add these values.

samggreenberg:
Short answer: **yes, this is useful — it's the missing third term of the decomposition the reports already lean on** — but the two "cheats" you named are not equally sound. The train-on-test SVM is a much more dangerous oracle than the threshold oracle it's modeled on, and the "halfway" option (full labels on the entire train split) is actually the principled primary, not the compromise.

## Why it's useful

Today `cost = oracle_cost + regret` (the vg-scale report's "Where the cost lives"). That split told us regret (the cut) is ~0.05 everywhere and doesn't separate the columns — "the ranking is where the remaining cost is." But `oracle_cost` conflates two very different causes:

- **the floor**: no linear head in this embedding space can separate the class, no matter the labels;
- **the shortfall**: a head *could*, but 10–200 clicks' worth of votes didn't find it.

Training regret splits exactly that:

```
cost = skyline_oracle_cost    (learnability floor: embedding + head capacity)
     + training_regret        (what the interactive loop left on the table)
     + regret                 (what the cut rule gives away on the ranking it got)  ← exists today
```

That answers the routing question no current metric answers: when a cell is bad, do you buy a better *embedder* or a better *acquisition/training loop*? vg-scale's `oracle_cost` spread (0.15–0.27 across columns, 0.36 at small size) is currently unattributable between those two. It would also sharpen the "stuck runs" story: a stuck run with a low skyline cost was a *findable* class the loop missed (acquisition problem); one with a high skyline cost was never learnable in that space (data/embedder problem). "Learnability" per (class, band, column) is exactly the skyline row's own oracle cost / AP.

## What you're missing — the sharp edges

[...]

Issue #3328, opened 2026-08-30
Report management


samggreenberg:
It's hard to browse experimental reports. Can we put the date in the folder name of the folders, so I can quickly see the latest experimental results?

samggreenberg:
Addressed in #3330 — every study directory is now `YYYY-MM-DD-<slug>` (dated by the day its report first landed), so `ls docs/experiments/` sorts chronologically. The index runs newest-first, the eight studies that had directories but no index row are listed, and a new `STUDY` invariant in `scripts/check-docs.py` fails the suite if a study lands undated or unlisted.

---
_Generated by [Claude Code](https://claude.ai/code)_

samggreenberg:
Shipped to `main` in the 2026-08-31 release — fixed in #3330.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3331, opened 2026-08-30
#3326: backfill the oracle line into all eight committed viewers, and vg_scale's supervised skyline

### LINKED ISSUE #3326: Re-analyse the committed studies so the viewer's oracle line and skyline notch have data
**Difficulty: light GRID work** — a re-analysis for most of it, one re-run for the skyline.

**Recommended Claude model: Sonnet 5.** The analysis commands are already written and the pass/fail signal is "does the notch appear"; the only judgment is which studies are worth the machine time.

Follow-up to #3325, which added two reference quantities to `scripts/experiments/calibration/viewer.py`:

- **the oracle-threshold line** (dotted) — the same momentary model cut where the test labels say it should have been cut; the gap to the solid performance line is the calibration regret;
- **the supervised-skyline notch** (right margin) — the #3322 learnability floor.

Both are wired end to end and covered by `selftest_viewer.py`, and every committed `viewer.html` has been re-skinned onto the new template. But **no committed study's payload carries either series**, so on all eight pages the oracle checkbox disables itself and no skyline notch appears. Neither can be recovered from the payload; both need the cell CSVs on the GRID.

The two are not the same amount of work:

### 1. The oracle line — a re-analysis, no re-run

`oracle_cost` / `oracle_fpr` / `oracle_fnr` / `n_test_pos` / `n_test_neg` are already on **every base row of every run ever done** — `viewer.py` reconstructs precision, recall and F1 at the oracle cut from them. So this is purely `viewer.py` against the existing `$CALIB_RESULTS`, no simulation:

```bash
python viewer.py --results "$CALIB_EXP" --arms results=prod \
  --baseline "$OUT/text_baseline.csv" --out "$OUT/viewer.html"
```

Then `scp` the page back over the committed one. Worth doing for every study whose results directory still exists:

- [ ] `docs/experiments/vg-scale/` — the one the request named
- [ ] `docs/experiments/good-mining-3267/`
- [ ] `docs/experiments/inclusion-knob-3196/`
- [ ] `docs/experiments/calibration-fraction-3287/` (+ its three per-column pages)
- [ ] `docs/experiments/calibration-fold-count-3310/`

Where a results directory has been reaped, say so on this issue and strike the row — the page keeps working, it just has no oracle.

### 2. The skyline notch — needs an actual re-run

`skyline_train_full` is vote-independent, so it is **one extra fit per arm per cell**, not one per click — cheap, but it has to be *run*: no existing results directory contains a skyline row.

```bash
CALIB_SKYLINE_ARMS=skyline_train_full ./launch_cells.sh
```

`vg_scale` is the one worth paying for: it is a 5-column map of an embedding space, and "how much of this cost is the space and how much is the loop?" is exactly what its five columns are being compared on. 3600 cells × one extra fit. Scoped to the whole-image columns in v1 — `siglip+dinov3_patch` is `max_patch` and the harness warns and skips there (open on #3321), so that column will have no notch.

### Checks

`viewer.py` prints a `NOTE:` naming what is missing when a build finds no skyline rows, and the page's own controls say why they are inert — so a page that shipped without either is loud, not silent. Confirm on each re-uploaded page that the oracle checkbox is live on cost/precision/recall/F1 and still disabled (with the "integrates over every threshold" tooltip) on AP and AUROC.

Once a page is re-uploaded, drop the "needs a re-run rather than a re-draw" paragraph from that study's `REPORT.md` — `vg-scale/REPORT.md` has one pointing here.

samggreenberg:
Closes #3326.

#3325 wired the oracle cut and the supervised-skyline notch end to end and re-skinned every committed `viewer.html` onto the new template — but a reskin carries whatever payload it finds. No page had either series, so on all eight the oracle checkbox disabled itself and no notch appeared. This backfills both.

## The oracle line — eight rebuilds, no new compute

`oracle_cost`, `oracle_fpr`, `oracle_fnr` and the split's class counts are on every base row this harness has ever written, and `viewer.py` reconstructs precision, recall and F1 at that cut from them. Every results directory still exists, so nothing was struck from the issue's list:

| page | results root | performance rows vs. the page it replaces |
|---|---|---|
| `vg-scale/` | `scale-3156-map` | identical |
| `good-mining-3267/` | `good-mining-3267` | identical |
| `inclusion-knob-3196/` | `incl-3196` | identical |
| `calibration-fraction-3287/` + `clip/`, `clip_l/`, `siglip2l/` | `calfrac-3287{,-clip,-clip_l,-s2l}` | identical |
| `calibration-fold-count-3310/` | `folds-3314/screen` | identical |

"Identical" is measured, not assumed: both payloads were decoded and compared, and `agg.mean`, `agg.sd`, `agg.n`, `agg.cells` and `runs.values` came back with `max|Δ| = 0` and matching NaN patterns on all eight. `good-mining-3267`'s `REPORT_generated.md` came back byte-identical for the same reason.

A wrong reconstruction draws a perfectly plausible dotted line, so the invariant was checked too: **the oracle cost is never worse than the achieved cost** at any of ~105,000 (group, arm, click) points across the eight pages, and click 0 — where there is no model — carries no oracle anywhere. (`fpr` and `fnr` individually *are* sometimes worse; the oracle minimises the weighted sum and trades one against the other. That is the quantity behaving.)

Two subtitles change, on the fold-count and calibration-fraction pages: #3325 rewrote those analyzers' subtitle text to name the overlay control, and a reskin cannot reach into a payload to update it.

## The skyline notch — a run, at a tenth of the budgeted cost

`vg_scale`'s floor comes from a **2880-cell pass over the four whole-image columns only**, into its own results root sharing the original grid's `prepare_info.json` and `crops` so the two agree on what each cell is. Two reasons not to re-run the 3600-cell grid:

* `siglip+dinov3_patch` is `max_patch`, where the v1 skyline is skipped (#3321) — and it is **89% of the grid's cost**. Running it buys one warning per cell.
* A re-run **replaces** the performance rows `vg-scale/REPORT.md`'s tables were read off, leaving a page that disagrees with the prose beside it.

The skyline is vote-independent, so a later pass over the same cells measures the same quantity. `viewer.py --skyline-results` merges it in while `--results` keeps supplying the curves; `selftest_viewer.py` plants the two roots and checks the rows come from the second one carrying the page's arm label, because a silent fallback would render as "this study has no floor" — indistinguishable on screen from the truth for a study that has none.

Result: 144 of 180 groups notched (36 categories × 4 whole-image columns, 20 seeds each), the pair column has none as intended, and no notch sits above the curve it belongs to. Mean floors order as the study would predict — `siglip2_l` 0.154 < `siglip` 0.188 < `clip_l` 0.199 < `clip` 0.217.

## Two things found on the way

* **`launch_scale.sh` reached its tasks with `CALIB_SKYLINE_ARMS` only through `--export=ALL`** — the one knob the launcher had not named explicitly, in a block whose comments explain why every other one is. Its failure mode is an array that runs the full 150-click loop and emits no floor, discovered hours later. Named it. (Preflight caught the fix while it was uncommitted, which is that check working.)
* **A CLI-built page now records its own build arguments.** `inclusion-knob-3196` ships chips called `linear` and `logistic` over results dirs called `svm` and `linear`, and which carried which was written down nowhere — it had to be settled by building both orders and diffing the decoded payloads. One reproduces the committed page exactly; the other differs by up to **0.45** in mean cost and would have inverted the study while looking entirely normal. `main()` now stamps the results root, the `dir=label` specs, the baseline and any `--skyline-results` into a `build` block that `--reskin` carries forward.

Also regenerated `good-mining-3267/startup_summary.json`, which was missing the `crossover` block `analyze_startup.py` now emits — purely additive, every other key byte-equal.

## Checks

`./run-tests.sh` on the GRID against this branch: **RUN PASSED** — 9326 passed, 6 skipped, 2 xfailed, all gates green (pyright, pip-audit, npm audit, Vitest). `check-docs.py` OK over 272 markdown files. `selftest_viewer.py` passes, including four new planted-answer checks for the split-root skyline and the build stamp. The 2880-cell array finished 2880/2880 COMPLETED with zero failures.

https://claude.ai/code/session_01R9nFHgJJCYGB7F5GNLyhvS


samggreenberg:
Merged `dev` in (#3330 / #3328, which dated every `docs/experiments/<study>/` directory).

GitHub reported the branch as `CONFLICTING`, but there was no content conflict — #3328 moved ~450 files and that is past GitHub's default rename-detection limit, so every one of my edited pages looked like a modify-vs-delete. Merging locally with `-c merge.renameLimit=999999` resolved it cleanly: git matched the renames and carried each rebuilt page onto its dated path.

Verified rather than assumed — `git diff --stat origin/dev HEAD` is exactly this PR's own 15 files and nothing else, all at the new paths:

- `2026-08-25-vg-scale/`, `2026-08-27-good-mining-3267/`, `2026-08-27-calibration-fraction-3287/` (+ its three sub-pages), `2026-08-28-calibration-fold-count-3310/`, `2026-08-29-inclusion-knob-3196/`.

Nothing I added references an undated study path.

Re-ran the suite on the merge commit: **RUN PASSED** — 9331 passed, 6 skipped, 2 xfailed, all gates green. That includes #3328's new `tests_lib/core/test_docs_gate.py` and `check-docs.py`'s seventh invariant, both of which pass over the rebuilt pages.

Now `MERGEABLE` / `CLEAN`.

Issue #3077, opened 2026-08-10
Make `clap_general` the default audio embedder; keep `clap` as the cheap tier and disambiguate their names


samggreenberg:
**Difficulty: low-medium** — the code change is small and mechanical; the care is in the fallout (defaults, docs, tests, and the display names users actually read).

**Recommended Claude model: Sonnet 5.** Bounded work against a spec that is already measured and decided. Step up to Opus 4.8 only if the default-embedder switch turns out to ripple further through the demo/cached-pickle paths than expected.

## What

Move `is_default` from `clap` (`laion/clap-htsat-unfused`) to `clap_general` (`laion/larger_clap_general`), keep `clap` available as an explicit cheap/fast option, and rename both display names so they can be told apart.

**Do not delete `clap`.** It is ~20% smaller and ~2.1× faster, which is a real reason to pick it, and existing dataset pickles and detector JSONs record `embedder: "clap"` — dropping it would stop those resolving.

## Why

Measured on the full ESC-50 (all 50 categories, 2000 clips) via `python -m vtscore.eval`, plus a threshold-free 1-NN probe. `clap_general` wins **every** comparison:

| | esc50_s | esc50_m | esc50_l |
|---|---|---|---|
| text-sort mAP — `clap` | 0.8634 | 0.8503 | 0.8661 |
| text-sort mAP — `clap_general` | **0.8950** | **0.8691** | **0.8842** |
| learned-sort mean F1 — `clap` | 0.5285 | 0.4572 | 0.5122 |
| learned-sort mean F1 — `clap_general` | **0.5641** | **0.5384** | **0.5230** |

Leave-one-out 1-NN over all 50 classes (400 clips): `clap_general` **0.973**, `clap` 0.958. (For context, `ast` 0.940 and `beats` 0.905 on the same probe.)

Cost, same probe: `clap` 77 s vs `clap_general` 162 s for 400 clips; 614 MB vs 776 MB of weights. `clap_general` is HTSAT 1024-d with depths `[2, 2, 12, 2]` against `clap`'s 768-d `[2, 2, 6, 2]`.

Note these numbers require the fix in #3076 — before it, the harness scored text queries with the media type's *default* embedder regardless of `--embedder`, so every non-default text-sort arm reported near-chance mAP.

## The naming is the real papercut

Today's display names are near-indistinguishable in the picker:

- `clap` → "CLAP (general audio)"
- `clap_general` → "CLAP (general 2024)"

[...]

PR #3340, opened 2026-08-31
#3127: keep enrich_descriptions off — it helps one default and costs another four times more

### LINKED ISSUE #3127: Re-check the `enrich_descriptions` default now that `clap_general` is the audio default
**Difficulty: low-medium** — the change itself is a one-line default flip; the work is the measurement that justifies (or rejects) it.

**Recommended Claude model: Sonnet 5.** Running `python -m vtscore.eval` arms and reading the numbers is bounded, mechanical work; no design judgment beyond "does the win hold across media types".

## What

Decide whether `enrich_descriptions` should still default to `False`, now that #3077 made `clap_general` the default audio embedder.

`enrich_descriptions` defaults to `False` today (`vtsearch/settings_models.py`, `ServerSettings.enrich_descriptions`). It is a single **global** setting — not per-embedder and not per-media-type — so its correct value depends on whichever embedders users actually run.

## Why now

Description enrichment flips sign between the two general CLAP checkpoints. From the ESC-50 sweep quoted in #3077:

| Embedder | Enrichment effect on text-sort mAP |
|---|---|
| `clap_general` | **+0.012 to +0.016** |
| `clap` | −0.005 to −0.014 |

Before #3077 the audio default was `clap`, where enrichment *hurts* — so `False` was the right global default for audio. After #3077 the audio default is `clap_general`, where enrichment *helps*, and the shipped default is now leaving that win on the table for every audio user who never opens Settings.

## Scope

- Re-measure enrichment on/off for the **default** embedder of each media type, not just audio — flipping a global on the strength of one media type's numbers is how you make image or video text-sort worse. `vtscore.eval` already has the text-sort mAP arm used for the #3077 numbers.
- If the win holds across the defaults, flip `enrich_descriptions` to `True` in `vtsearch/settings_models.py` and note it in `CHANGELOG.md`.
- If it does not hold, the setting is being asked to do two incompatible jobs; say so and close this, or spin out "make enrichment per-embedder" as its own issue.
- Either way, `docs/ML.md`'s description of enrichment should end up stating which embedders it helps, so the next default move doesn't silently re-open this.

Follow-up to #3077 (its "Watch out for → Enrichment flips sign between the two" note).

samggreenberg:
Closes #3127.

## Verdict: keep `enrich_descriptions = False`

#3077 moved the audio default to `clap_general`, where enrichment *helps*, and
asked whether the global default should follow. It should not. Enrichment helps
exactly one of the four media-type defaults, by less than this study can
resolve, and costs the text default four times more than it wins there.

| media type | default | mAP plain | mAP enriched | Δ AP (paired) |
|---|---|---|---|---|
| audio | `clap_general` | 0.88 | 0.90 | **+0.014 ± 0.009** |
| image | `siglip` | 0.59 | 0.59 | −0.001 ± 0.002 |
| text | `e5` | 0.61 | 0.55 | **−0.057 ± 0.009** |
| video | `xclip` | 0.82 | 0.83 | +0.008 ± 0.014 |

Equal-weighted over the four defaults the setting is negative (−0.009 ± 0.005).
The rule pre-registered in [`PLAN.md`](../blob/claude/enrich-descriptions-3127/docs/experiments/2026-08-31-enrich-descriptions-3127/PLAN.md)
before the run returns `DO_NOT_FLIP__HARMS_A_DEFAULT`.

**Full report:** [`docs/experiments/2026-08-31-enrich-descriptions-3127/REPORT.md`](../blob/claude/enrich-descriptions-3127/docs/experiments/2026-08-31-enrich-descriptions-3127/REPORT.md)

## Three findings worth more than the verdict

1. **#3077's premise reproduces exactly — and is smaller than it looked.** The
   sizing cell returned `clap_general` esc50_s mAP 0.8950, the published number
   to four decimals, and the sign flip holds (`clap_general` +0.014, `clap`
   −0.010). But neither clears twice its own clustered SE. The win that
   motivated the issue is a point estimate, not a resolved effect.
2. **The audio win is one template, not an ensemble.** `the sound of {text}`
   alone is worth +0.014 on `clap_general`; the other four are +0.002, +0.002,
   −0.004 and (the identity) 0.000, and the five-way mean lands exactly where
   the single template already was.
3. **Every wrapper hurts `siglip`, `e5`, `bge` and `clap`.** On those four there
   is no member of the set that beats the typed text. The ensemble only looks
   harmless because `{text}` is itself one of the five and averages the plain
   query back in — which is why `siglip`'s −0.001 is not "enrichment does
   nothing to images" but four negative templates diluted by one identity.

The text result is not one checkpoint: 45 of 45 categories worse on `e5`, and
45 of 45 worse on `bge`. The obvious mechanism — E5's asymmetric
`query:`/`passage:` convention — was tested with a probe and **refuted**:
crossing the entire asymmetry gap costs −0.015 ± 0.008 against enrichment's
−0.057. Three quarters of the damage is unexplained, which is itself a reason a
*global* default should not move.

## What's in the diff

**No behaviour change.** The setting keeps its default and its code path.

- `docs/experiments/2026-08-31-enrich-descriptions-3127/` — plan, report,
  reading copy, 13 tables, 4 figures, machine-checked `verdict.json`.
- `scripts/experiments/enrich/` — runner, launcher, analyzer (the decision rule
  is module constants, so the verdict is computed rather than written),
  figure module with its own CLI, and the mechanism probe.
- `docs/ML.md` — a **Description enrichment** section; it had none, and #3127
  asked for one that says which embedders it helps.
- `frontend/.../settings-modal.component.html` — **the tooltip was wrong.** It
  read *"Prepend item filenames to text-sort queries to improve matching for
  named items"*, which is a different feature; the setting has never touched
  filenames. Found by this study, fixed with it. `CHANGELOG.md` notes it.
- `tests/sorting/test_enrich_descriptions.py` — `test_default_is_false` now
  says *why*, with the numbers and a pointer to the report.
- **Drive-by:** `docs/experiments/2026-08-30-fit-quality-3329/REPORT.md` linked
  into `docs/reports/`, which the #3336 restructure removed — `check-docs` was
  red on `dev` before this PR and is green after.

## How it was run

28 cells on the GRID (4 chunks + 2 control arms + 1 probe), ~40 min of A100
time, all `COMPLETED`, no zero-byte cells, peak RSS 11.4 GB. One dataset load
serves every arm, so plain vs enriched differ in the query vector and nothing
else — same medias, same encoder, same card, same process. The `{text}` wrapper
is an identity and its solo arm reproduces the plain arm to `max |Δ| = 0.0` on
all six embedders; that planted answer is checked before any verdict is read off
the same rows.

## Follow-up

Filed separately: make enrichment **per-embedder**. Only two templates in the
tree have a positive point estimate on their own embedder, and four of the six
embedders measured have none that beats the typed query at all.


samggreenberg:
Full suite green on `9e678514` (GRID cpu node, 8m41s): **9375 passed**, 6 skipped, 2 xfailed; pyright, pip-audit, npm audit and Vitest all OK.

Note the drive-by was load-bearing, not cosmetic: `check-docs` runs *before* pytest in `run-tests.sh`, so the stale `docs/reports/` link left by the #3336 restructure was failing the whole suite on `dev`. It is green either side of that fix now.

Issue #3319, opened 2026-08-29
Extend the acquisition-offset grid past -4, and check the deep regime


samggreenberg:
**Difficulty:** low-medium — the harness, the launcher and the analyzer all exist and have just run this grid. What is new is two arms and a longer horizon.

**Recommended Claude model: Sonnet 5.** The mechanics are a re-launch of #2877's grid with a wider arm table. Step up only if the answer comes back non-monotone, which would be a shape finding rather than a parameter one.

## What

Extend the acquisition-offset grid past `-4` on the arm that ships, and check the deep regime.

PR #3319 shipped `ACQUISITION_INCLUSION_OFFSET = -3` on the evidence in [`REPORT_PILE_2877.md`](../blob/dev/docs/experiments/acquisition-inclusion/REPORT_PILE_2877.md). On the shipped default arm (`vg_scale_any × siglip × whole_image`, 192 paired cells), measured against the previously shipped `-1`:

| arm | Δ final cost vs `-1` | Δ positives@100 | Δ AP | deep spikes |
|---|---:|---:|---:|---|
| `-2` | −0.013 [−0.026, −0.012] | +4.5 | +0.020 | 0.0% → 1.0% |
| `-3` | **−0.020** [−0.029, −0.011] | +10 | +0.045 | 0.0% → 0.0% |
| `-4` | −0.017 [−0.030, −0.013] | **+18** | **+0.057** | 0.0% → 0.5% |

**`-4` is at least as good as `-3` on every endpoint and the trend has not turned.** So `-4` is the edge of the grid, not an optimum, and the shipped `-3` is a *conservative* choice made because `-4` costs a small resolvable regression under region voting (+0.006, CI [+0.001, +0.013]) where `-3` is free.

Two things follow, and neither is answerable from the existing cells.

## 1. Where does the frontier actually turn?

Run `-5`, `-6`, `-8` on the shipped arm. Three outcomes:

- **It turns between −4 and −6.** Then there is a real interior optimum and `-3` is leaving measurable quality on the table; ship the optimum.
- **It keeps improving.** Then the mechanism is not "aggression has a price" but something closer to "the reporting cut is simply mis-placed for acquisition", and the right object is not an offset at all — it is a separate acquisition rule, which changes the shape of the knob rather than its value.
- **It degrades sharply somewhere.** Then `-3` is confirmed as a safe interior point, which is worth knowing with a bound rather than by assumption.

[...]

Issue #3166, opened 2026-08-17
The GMM cut is nondeterministic on a saturated score distribution


samggreenberg:
**Difficulty: small-to-medium.** **Recommended Claude model: Sonnet 5** — the reproduction is mechanical, but deciding *which* determinism fix is right (seed, init, or refusing to fit) needs a little judgement about the threshold path.

## What happened

Re-running a finished benchmark cell with a longer vote horizon reproduces the original run **bit-for-bit** on every overlapping step — 52,717 shared steps across three grids — with exactly one exception:

`caltech101_m` / `siglip` / `airplanes` / seed 1, from vote 37 onward:

| column | max |difference| |
|---|---|
| `n_good`, `n_bad`, `average_precision`, `auroc` | **0** |
| `threshold` | **0.0264** |
| `cost`, `fpr` | 0.0195 |

Identical votes, identical training set, identical ranking — and a different cut. So the divergence is entirely inside threshold selection, and it is machine-dependent (the two runs differ only in which node they landed on).

## Why that category

`airplanes` on `caltech101_m` is saturated: positives score near 1.0, negatives near 0, and nothing sits between them. The mixture fit that places the cut is ill-conditioned on a distribution like that — there is no interior structure to find — so a float-level difference in BLAS ordering or thread scheduling is enough to land it in a different local optimum. Every non-saturated cell in the same three grids reproduced exactly, which is consistent with that reading.

## Why it matters beyond one cell

- **An experiment cannot distinguish this from a real effect.** In the run that found it, the apparent "cost rose by +0.005 ± 0.002 when we clicked longer" on `caltech101_m` was entirely this instability. It was caught only because that study had a bit-identity check between two runs of the same cells; a study without one would have reported it.
- It is a **shipped path**, not an eval artefact: the same cut is what a user's detector gets.
- The impact is bounded — a saturated distribution is one where any cut in a wide interval performs identically, which is why cost moved only 0.02 — but "the threshold is unreproducible" is a bad property for a value we calibrate, sweep and regress-test.

## Suggested work

1. **Reproduce synthetically**: two well-separated score clusters, no interior mass, fit the cut repeatedly under different thread counts (`OMP_NUM_THREADS`) and show the threshold moving. That is a unit test.
2. **Decide the fix.** Options, roughly in order of preference:
   - seed the mixture fit explicitly (`random_state`) so it is at least deterministic per input;
[...]

Issue #2836, opened 2026-08-04
The GMM cut optimizes a count loss while we score a rate loss: derive the right cut instead of reverting to the midpoint


samggreenberg:
**Recommended Claude model: Opus 4.8.** Mostly theory + experiment design on an existing harness; the risk is a plausible-but-wrong derivation, which is exactly where a weaker model is expensive.

## The puzzle

#2799 measured the equal-density crossing cut (#2798, shipped in #2801) against the historical midpoint-of-means, as paired within-step variants (each re-cuts the *same* model on the *same* votes). The crossing lost:

| geometry | window | Δcost (crossing − midpoint) | ΔFPR | ΔFNR | p |
|---|---|---|---|---|---|
| **pooled** (production, post-#2797) | 6–20 votes | **+0.0036** | −0.0114 | +0.0150 | 1.9e−08 |
| **pooled** | 2–5 votes | **+0.0059** | −0.0234 | +0.0293 | 4.4e−05 |
| image-level (pre-#2797 geometry) | 6–20 votes | **−0.0022** | −0.0041 | +0.0019 | 1.6e−04 |
| image-level | 2–5 votes | −0.0018 | −0.0012 | −0.0006 | 0.19 (n.s.) |

Note the **inversion**: the crossing *helps* in the image-level geometry and *hurts* in the max-pooled one — the exact geometry #2798's argument was about. The theory predicted the crossing's direction correctly (it does sit above the midpoint, buying FPR at the cost of FNR) and predicted it would help most where max-pooling fattens the Bad mode. It overshoots most in precisely that place.

The empirical answer is "revert to the midpoint" (#2833). That is unsatisfying: the midpoint has no theory behind it, so shipping it means we are choosing a cut we believe is wrong because it scored better. This issue is to find out **why** it scored better, on the expectation that the repaired theory names a third point that beats both.

## Leading hypothesis: we are solving for the wrong loss

`_weighted_gaussian_crossing` (`vtscore/training/thresholds.py`) solves

```
w_lo · N(x; mu_lo, var_lo)  ==  w_hi · N(x; mu_hi, var_hi)
```

That is the Bayes boundary **with the mixture weights as class priors** — it minimizes expected misclassification *count*.

What we score, and what the Inclusion knob is defined in terms of, is

```
cost = fpr_weight * FPR + fnr_weight * FNR      # inclusion_weights(0) == (1.0, 1.0)
```

[...]

PR #3454, opened 2026-09-01
Extend the acquisition-offset grid past -4 and to half steps: a plateau, not a peak (#3319)

samggreenberg:
Answers the first two questions of #3319 on the shipped arm. **2304/2304 cells, 0 failures**, 12 arms × 192 paired cells on `vg_scale_any × siglip × whole_image` — #2877's environment verbatim, every incumbent re-run on one dev commit. The 400-click deep wave and the region cross-check are in flight and will be pushed to this branch.

The falsifier behaved (`k=+2`: positives 7 → 4, cost +0.063 [+0.053, +0.073]), so the run is interpretable.

## 1. The frontier does not turn where anyone was looking

`final_cost` is **statistically flat from −2 to −5** — a three-bit-wide plateau — and only degrades resolvably at **−8** (+0.024 [+0.015, +0.032] against the best arm). `−6` is +0.011 [+0.003, +0.019]: clears zero, not the ±0.010 tolerance. Everything from −3 to −5 sits within [−0.006, +0.011] of the minimum with every CI spanning zero.

**So the shipped `−3` is mid-plateau: confirmed safe, and confirmed arbitrary.**

What is *not* flat is the mechanism — positives per 100 clicks rise monotonically and without saturation across the whole grid (7 → 20 at −3 → 38 at −5 → **64 at −8**), AP 0.568 → 0.722. What stops the sweep is the **guardrail**, not the cost: deep-spike incidence 0–0.5% out to −5, then 1.0% at −6 and **2.6% at −8**. The same criterion #3318 found binding.

This is closest to the issue's **second outcome**: cost is not pricing aggression in this range, it is saturated.

## 2. Half steps are real, and decision-irrelevant

They could have been an artefact — `threshold_at` snaps its realised quantile to the haystack sample (#3166), so a half step might have collapsed onto its integer neighbour and been a *silent duplicate*. It does not: across **2112 adjacent-arm cell comparisons, zero** share an identical `acq_pool_percentile`.

They are also genuinely half-way. All six **mechanism** contrasts resolve (+2.9 to +5.1 positives above the shallower neighbour, −4.2 to −4.9 below the deeper one, every CI excluding zero); **none** of the six **cost** contrasts does.

H2 is falsified as pre-registered, but not for the anticipated reason. The knob is not too coarse to have half-step resolution — it has it. The decision endpoint cannot see half a bit **because it cannot see three whole bits either**. Retired as a tuning device.

## 3. What "inclusion −3" means, and a measured ~2-bit calibration debt

Because `inclusion_cost_weights` is a weighted sum of *rates*, prevalence divides out (`rate_crossing` puts the prior-odds factor back into `lam` precisely so the cut does not carry it — #2836). What remains is a **log₂ likelihood-ratio threshold**: include *x* iff `f_pos/f_neg > 2^−k`. **Each step is one bit of evidence**; `−3` asks for 8:1.

Tested on the **pick log**, where the claim belongs (17.6k `hard` picks per arm):

| k | 0 | −2 | −3 | −4 | −5 | −6 | −8 |
|---|---|---|---|---|---|---|---|
| hard-pick precision | 4.7% | 15.7% | 24.0% | 34.8% | 44.5% | **52.9%** | 60.9% |

The **shape is exactly right** — smooth and monotone over eleven arms, which is what a log-odds threshold should produce. At `k=0` the hard pick returns **4.7%, below the 7.1% base rate**: sampling at the decision boundary is worse than random, the cleanest statement of why this offset exists.

The **location is wrong by a constant**. Precision crosses the posterior-flip point (`k* = −log₂((1−π)/π) = −3.71` at π=7.1%) at **k ≈ −5.66**, so the fitted mixture **overstates the evidence by ≈3.9×**. That is a result, not a failed hypothesis: the machinery is correctly specified and mis-calibrated, and the gap is now measured in the knob's own units. It points at a knob shaped like a **target pick precision** rather than a constant offset.

## Ship status — deferred, deliberately

Four arms pass against the incumbent `−3` (`−3.5`, `−4`, `−4.5`, `−5`); H2 says pick an integer, so `−4` or `−5`. **No recommendation is made from this half**: #3318 rejected `−4` under region voting where `−3` was free, and the region cross-check on the survivors is running. Deferring is what the plan pre-registered.

## Code

The arithmetic was always continuous (`2**-k` throughout), so accepting fractional offsets is annotations plus one env parser, not a behaviour change; production keeps the integer `−3`.

- `thresholds.py` / `voting_iterations.py` — widen the inclusion annotations to float, and document the likelihood-ratio reading at the single definition.
- `experiment_config.py` — parse `CALIB_ACQ_INCLUSION_OFFSET` as a float. As an int, `-3.5` raises at import and kills the arm at cell 0; a silent *truncation* would be far worse — the arm runs, completes, and is a duplicate of `−3`.
- `analyze_acq.py` — derive `ARM_K`/`ARM_LABEL` from the arm name (`h` = half step), arm list from `ACQ_ANALYZE_ARMS`, defaulting to #2877's seven so that run still reproduces.
- `frontier_3319.py` — the shape questions, which are not a by-product of the per-arm table. Reads the turn off CIs, never a fitted parabola (that manufactures an interior optimum whether or not one exists), and refuses an arm whose pool percentile matches a neighbour cell-for-cell.
- 4 new tests pinning the bits semantics, fractional offsets, and that half steps land strictly between their integer neighbours.

**Power, stated rather than rounded away:** realised paired SD 0.0747 needs n=215 for ±0.010; the run has 192, giving ±0.0106. It changes no verdict — the plateau's contrasts are an order of magnitude inside the tolerance and the `−8` rejection far outside it. Resolving two adjacent plateau arms on cost would need ~5400 cells/arm, which is the honest reason the half-step grid is retired rather than re-run bigger.

https://claude.ai/code/session_01VZCCqYyXjP8d48nmPtRBYu


samggreenberg:
## Complete — deep wave and region cross-check are in

Three waves, **4032 cells, 0 failures, 0 header-only cells**.

### Ship `-4`

The only arm passing the pre-registered rule in **all three** environments. Against the incumbent `-3`:

| environment | Δ final cost [95% CI] | Δ positives | Δ AP | deep spikes | passes |
|---|---|---:|---:|---|---|
| 100 clicks, `siglip × whole_image` | −0.0015 [−0.0081, +0.0050] | +9.9 | +0.019 | 0.0% → 0.5% | YES |
| 400 clicks, `siglip × whole_image` | +0.0004 [−0.0029, +0.0037] | +9.7 | +0.004 | 5.7% → 2.1% | YES |
| region, `pair × max_patch` | +0.0031 [−0.0025, **+0.0091**] | +9.4 | −0.002 | 1.6% → 1.6% | YES |

`-5` is free on the shipped arm and **fails region** (+0.0064 [+0.0007, +0.0123] against a +0.010 bar) — exactly the constraint #3318 raised.

**The region margin is thin, stated rather than rounded away.** #3318 measured this same contrast at +0.006 [+0.001, +0.013] and rejected it; this run measures +0.0031 [−0.0025, +0.0091] and passes. The CIs overlap heavily — consistent measurements straddling the bar, not a reversal. **The positives case (+9.4 to +9.9 per 100 clicks, three environments) is much stronger than the cost case, which is a null in all three.** Anyone uncomfortable with a decision turning on 0.0009 of cost should treat `-3` → `-4` as optional.

### H3 — the deep regime does not flip the sign

The falsification condition was that the 400-click optimum be *shallower*. It is not: `-3` and `-4` are tied at both horizons, and in absolute terms the offset is worth **more** at depth (paired cost −0.033 vs −0.031; positives **+90** vs +17.7). The issue expected the benefit to fade where scarcity ends; the likelihood-ratio reading predicted the opposite, since the selector ranks the *unvoted* pool whose prevalence falls as positives are harvested.

**What is new at depth is the guardrail.** Deep spikes were 0.0% for every arm out to `-5` at 100 clicks; at 400 clicks `-3` goes 0.5% → **5.7% (p=0.006)**. Non-monotone (`-4` is 2.1%), so recorded as a live hazard, not as evidence about which arm is safer.

### The calibration gap replicates — and tracks how well the ranking separates

| environment | 50% crossing | predicted `k*` | gap | overstated by |
|---|---:|---:|---:|---:|
| binary — `siglip × whole_image` | −5.66 | −3.71 | **1.95 bits** | 3.9× |
| REGION — `pair × max_patch` | −4.24 | −3.71 | **0.53 bits** | 1.4× |

The gap is **not a constant of the codebase** — it measures how much the fitted mixture overstates its own evidence, and it **shrinks when the ranking separates well**. Region voting takes oracle cost 0.382 → 0.218 on identical cells (#3318), so its nominal bits are closer to true bits.

That one quantity retro-explains the constant's messy history: why the optimum moved between environments, why region voting tolerates aggression binary voting does not, and why one global value has always needed re-measuring. **The arms were never sweeping "aggression" — they were sweeping *nominal* bits against an environment-dependent calibration debt.**

### A correction, and the check it produced

The plan named positive exhaustion as the artefact that would fake exactly this result — and it was cleared from **one sizing cell** (`backpack` seed 0: 57 of ~150, rate still rising). The full wave says otherwise:

| arm | median positives @400 | median harvest of the sim half | cells >90% harvested |
|---|---:|---:|---:|
| `prod` | 22 | 14.7% | 0.0% |
| `acq_m3` | 123 | **82.0%** | 21.9% |
| `acq_m4` | 128 | **85.3%** | 29.2% |

The report sentence claiming the tail was not positive-limited is **retracted in place**. The confound is one-sided — the aggressive arms hit a ceiling the control never approaches, which can only *compress* their advantage — so the win survives it. But it costs the deep wave the other half of its question: ceiling-limited arms cannot say whether the optimum gets *deeper*.

**`preflight.sh` check 16b** now catches this. Check 16 asks whether the horizon empties the sim set of *media* — which the 400-step run passed comfortably on 3873 media while harvesting 85% of the 150 positives inside them. 16b reads `category_counts`, takes the thinnest selected category, and notes when `max_steps` reaches the positives available. Verified to fire at 400 steps and stay quiet at 100.

Lesson filed: **a pilot answers questions about the MACHINE, never about the DATA.** Timing and memory generalise from one cell; positives-per-category is the most category-dependent quantity in this harness.

samggreenberg:
**Full suite green on the GRID** (job 609060, worktree at `3cedb1173`): `9485 passed, 6 skipped, 2 xfailed` in 7m17s; pyright, pip-audit, npm audit, Vitest, deptry, the docs gate and the OpenAPI snapshot all OK. `RUN_TESTS_EXIT=0`.

samggreenberg:
## The endpoint was the wrong lens — the offset buys 2–3× *speed*

`final_cost` is one point on a curve, and every report on this constant (this one included, until now) read the knob through that point alone. Added two paired trajectory endpoints — **clicks-to-target** (first crossing of that cell's own control final cost, measured identically on both sides) and **area under the cost curve**.

**Median clicks to reach the answer `k=0` ends its session with:**

| | `k=0` | `k=−1` | `k=−3` (shipped) | `k=−4` |
|---|---:|---:|---:|---:|
| 100-click session | 47.5 | 30.5 | **23.5** | 25.0 |
| 400-click session | 210.5 | 137.0 | **80.0** | **65.5** |

**~2× fewer clicks over a short session, 3.2× over a long one — and it compounds.** Paired against `−3`, running with no offset costs **+17.1 clicks [+12.0, +21.8]** at 100 and **+101.1 [+82.7, +119.5]** at 400. AUC agrees (`−3` vs prod: −0.0217 [−0.0305, −0.0131] at 100; −0.0311 [−0.0369, −0.0256] at 400), as does the symmetric matched-budget view: **`−3` at 200 clicks is already better than `prod` at 400.**

### The plateau survives the new endpoints

Which is what makes it a finding rather than an artefact of one metric. The speed measures draw the plateau's **edges** more sharply than cost did — `prod`, `−1`, `−2` and `−8` are all resolvably worse than `−3`, where on cost `−2` was not — but inside `−2.5 … −6` every contrast is a null on both. At 400 clicks, `−4` vs `−3` is ΔAUC −0.0018 [−0.0053, +0.0016] and −4.4 clicks [−15.8, +7.1].

**So `−3` vs `−4` is now a null on three endpoints at two horizons.** The ship case for `−4` rests on **labelling efficiency alone** — not quality, not speed. The report says so plainly.

Under region voting the speed gain is real but much smaller (45.5 → 32 clicks) with AUC a null and 27–35% of cells never reaching target — the same story as #3318's DiD, and the calibration section now explains why.

### Two honesty notes, carried in the text

- The per-cell crossing median is **not** the crossing of the median curve (23.5/80 vs ~41/145). Both agree in direction; the per-cell one is properly paired, and the figure shows both.
- `prod` reaches its own target in 100% of cells by construction while the arms miss in 8–22%, so arm medians are conditioned on getting there.

### Also: the calibration characterisation is corrected

Regressing log₂-odds of pick precision on `k`: intercept **−3.79**, slope **−0.65** (binary); −3.32 / −0.78 (region), against a calibrated −3.71 / −1.00. **The origin is right and the steps are short** — a different defect from a constant offset, with a different fix. But it is not a constant gain either: residuals curve, and `acq_p2` (`k=+2`), which was not in the fit, lands closer to the *calibrated* line than the fitted one on the binary arm. "Approximately calibrated near 0, increasingly compressed with depth, saturating" is what the data supports; the earlier "3.9× overstated" was the accumulated shortfall at the crossing, not a multiplier.

### Report restructured

Rewritten as **one document** rather than a report with a speed section appended. The trajectory result reframes the study — it is what the knob is *for* — so it leads; the plateau section now carries cost, AUC and speed together; and the ship section states which endpoints the recommendation does and does not rest on. `check-docs` OK (291 files).

samggreenberg:
## Shipped `-4`, and merged up to dev (now 0 behind)

### The constant

`ACQUISITION_INCLUSION_OFFSET = -4`, with the rationale block rewritten to say **what the case actually is**: the decision endpoint is a plateau and `-3` vs `-4` is a null on final cost, AUC *and* clicks-to-target at both horizons, so this is **labelling efficiency** (hard-pick precision 24% → 35%), not quality and not speed. The thin region margin (+0.0031 [−0.0025, **+0.0091**] against a +0.010 bar, where #3318 measured +0.006 [+0.001, +0.013] and rejected) is recorded in the constant's own docstring as *the number to re-measure first* if the value is ever revisited.

Sites: the constant; `inclusion_cost_weights`'s worked example (8:1 → 16:1); the `simulate_voting_iterations` docstring; three test assertions; the haystack-sizing note in `test_acquisition_threshold.py`; `docs/ML.md`.

### Three accretion bugs fixed en route

All the same shape — **a value restated in prose beside the constant, left behind when the constant moved**:

1. The `thresholds.py` rationale block **opened** with *"`-1` is the only value that passes … in all four environments"* while shipping `-3` (left from #2909).
2. `docs/ML.md` opened *"The offset is −3"* and closed *"keeps −1 global … −1 is the only value passing in all three"*, with a later paragraph still ending *"`-1` remains the best-supported single global value"*.
3. `experiment_config.py` said **"Do not restate the value here"** and then restated it — *"it is **-1** today"* — three ships out of date.

### The merge: 173 commits, one structural conflict

**dev split `vtscore/training/thresholds.py` into a five-module package** (7f31b9d97) while this branch was editing it — a modify/delete, with no text to merge. Every edit was re-applied against an asserted anchor in the new layout (`knobs.py`, `anchored.py`, `conformal.py`). `docs/ML.md` had one hunk where dev corrected the module path in the same sentence this branch changed the value; took dev's path with this branch's value.

**Verified by name, not by the merge exiting 0.** `experiment_config.py`, `preflight.sh` and `voting_iterations.py` all auto-merged *cleanly*, and a clean auto-merge that drops a hunk looks identical to one that keeps it. A 33-assertion check confirms every #3319 change is present in the merged tree. It found nothing lost — and one false negative of its own (a needle spanning a `#:` line-wrap), confirmed present by hand rather than assumed.

**Two follow-on fixes the merge required**, both artefacts of dev refactoring *after* this branch widened the callers:

- `scripts/check-calibration-index.py` (a gate new in dev, #3409) blocked the suite because the two new scripts were unindexed.
- **pyright** found a float offset flowing into two int-annotated parameters — `_resolve_run_knobs` and `threshold_from_fold_orderings`, both extracted by dev after the widening. Real, and exactly what a clean merge hides.

### Suite status

**9781 passed**, 6 skipped, 2 xfailed. pyright OK, pip-audit OK, vulture OK, Vitest OK, deptry OK, docs gate OK, calibration index OK.

**One gate is red and it is not from this branch: `npm audit`** — a moderate `qs` advisory. This branch touches **zero** frontend or dependency files (all 23 changed files are Python, Markdown, shell or PNG), and `frontend/package-lock.json` is **byte-identical to dev** — `git diff origin/dev -- frontend/package-lock.json` is empty, and auditing the lockfile directly reproduces the advisory. So dev is red on this gate too. Fixing it means `npm audit fix` mutating the lockfile, which is a dependency change unrelated to a calibration constant — **flagging rather than folding it into this PR**; it should be its own issue.

Issue #3547, opened 2026-09-02
Long labelling sessions are untuned, and the pile cannot currently measure them


samggreenberg:
**Difficulty:** low to build, medium to interpret — the blocker is a *dataset* property, not harness work.

**Recommended Claude model: Sonnet 5** for the pile rebuild; step up if the deep result comes back disagreeing with the 100-click one, which would be a shape finding.

## What

**Every value of `ACQUISITION_INCLUSION_OFFSET` ever shipped was chosen at a 100-click horizon** — #2876, #2877, #2905, #2909, #3318 and now #3319's headline. #3319 ran the first deep wave (400 clicks, 768 cells) and found two things: the good news held, and the environment ran out of room to answer the rest.

### The guardrail wakes at depth

Deep-spike incidence is **0.0% for every arm out to `−5` at 100 clicks**. At 400 clicks:

| arm | k | deep spikes vs `prod` |
|---|---:|---|
| `acq_m1` | −1 | 0.5% → 1.6% (p=0.63) |
| `acq_m3` | −3 | 0.5% → **5.7% (p=0.006)** |
| `acq_m4` | −4 | 0.5% → 2.1% (p=0.38) |

That is the **first guardrail cost this constant has shown anywhere on the shipped arm**. It is non-monotone — `−4` is *lower* than `−3` while sampling more aggressively — which is what a low-rate count (11 vs 4 events in 192 cells) looks like when it is noisy, so #3319 recorded it as a hazard rather than as evidence about which arm is safer. **It has not been resolved, and it sits in the regime real users occupy on a long labelling session.**

### The environment cannot answer the deeper question

At 400 clicks the aggressive arms hit their **positive ceiling** and the control never comes near it:

| arm | median positives @400 | median harvest of the ~150-positive sim half | cells >90% harvested |
|---|---:|---:|---:|
| `prod` | 22 | 14.7% | 0.0% |
| `acq_m1` | 36 | 24.0% | 0.0% |
| `acq_m3` | 123 | **82.0%** | 21.9% |
| `acq_m4` | 128 | **85.3%** | 29.2% |

[...]

Issue #2910, opened 2026-08-08
Make the acquisition offset supply-dependent instead of constant — recover COCO's 3x without charging the other environments' tails


samggreenberg:
**Difficulty:** medium-hard — the harness knob, the launcher and the analyzer all exist and have run three times. What is new is that the offset stops being a constant, which means the app path and the eval path have to agree about a value that now *moves during a session*.

**Recommended Claude model: Opus 5.** The mechanics are bounded; the risk is that the obvious way to measure this is contaminated, and the last study nearly shipped exactly that mistake.

## What

Replace the constant `ACQUISITION_INCLUSION_OFFSET` with a **supply-dependent** one: aggressive while the detector is starved of positives, relaxing toward 0 as they accumulate.

## Why

Three environments have now measured the constant and they do not agree (#2876, #2877/PR #2891, #2905/PR #2909):

| environment | voting | `-3` |
|---|---|---|
| `coco_val × siglip2` | binary | ships — positives 4 → 18, cost −0.011 |
| `visual_genome_m × siglip` | binary | **fails** — cost CI [+0.003, +0.022] |
| `visual_genome_m × dinov3_patch` | region | passes, but the mechanism is absent |

#2909 shipped `-1` because it is the only value passing everywhere. **That is a compromise, not an optimum, and it is an expensive one**: on a starved COCO-like environment `-1` finds 6 positives per 100 votes where `-3` finds 18. We gave up a 3× effect to avoid a regression in a different environment.

**The measurements say the two environments want different things for a reason we can observe at run time.** Under binary voting the offset's benefit is sharply concentrated in *starved* cells and turns **negative** in well-supplied ones. Measured on axes independent of the arm being scored (this matters — see below), the AP response to `k=-3` has slope **−0.0207** on log category prevalence (CI [−0.0259, −0.0159]) and **−0.0402** on a leave-one-out baseline. At the most starved quintile `k=-3` moves AP **+0.0814** and oracle cost **−0.0253** — it helps *globally*. At the richest, AP +0.000 and oracle cost **+0.0356**.

So the offset is a **starvation remedy whose price is charged everywhere**. A constant has to pick one point on that curve for every user. The detector already knows where on the curve it is: it knows how many positives it has.

[...]

Issue #3414, opened 2026-09-01
The Smart-indicator FP/FN cost loop is a mirror that doesn&#39;t need to be one


samggreenberg:
**Difficulty: S.**
**Recommended Claude model: Opus 4.8.** It touches the shipped stopping-advice and every study's vote order simultaneously.

`vtscore/eval/voting_iterations.py:2826` (medium impact) **[re-pin]**

`_labelset_error_costs` (2826-2839) and `labeling_progress._score_step` (`vtscore/detectors/labeling_progress.py:691-699`) copy the same weighted-error arithmetic — pure math with none of the can't-delegate reasons the sync-gate docstring lists. Also: `voting_iterations._inclusion_weights` (685) and `calibration_metrics.inclusion_weights` are two wrappers in one package over the same production function.

## Direction

Add `weighted_error_cost(scores, labels, threshold, fpr_weight, fnr_weight)` to `vtscore/training/thresholds.py`; both callers delegate (each keeps its own plumbing). Delete `_inclusion_weights` in favour of the `calibration_metrics` import it already has. Coordinates with the August audit's `_eval_cached_models` delegation item — ship together.

## Constraints

This moves logic pinned by `scripts/check-eval-app-sync.py`. The mirror whose app side is `labeling_progress._score_step` and whose harness side is `_labelset_error_costs` in `vtscore/eval/voting_iterations.py` becomes a delegation once both callers route through the new `vtscore/training/thresholds.weighted_error_cost` — so the `Mirror(...)` entry should either be re-pointed at the surviving thin call sites or retired outright if there is no longer a hand copy to guard. Whichever you choose, run:

```
python scripts/check-eval-app-sync.py --update
```

after reconciling, never before. If any deliberate difference survives between the two call sites, record it in that mirror's `divergence=` text so the next person reconciling knows it is intentional.

---
Surfaced by the September 2026 structure audit (`docs/plans/codebase-audit-2026-09.md`).

samggreenberg:
Addressed in #3568.

Three corrections to this issue's premise, recorded here because they changed the shape of the fix:

[...]

PR #3598, opened 2026-09-03
#3547: the frontier analyzer, and a sizing amendment the first cells forced

samggreenberg:
Follows PR #3584 (merged), which built the `vg_scale_deep` pile. A fresh PR because that one auto-merged out from under the branch.

The 1344-cell grid is **running** (7 arrays, ~11:15 EDT). This adds the analyzer it needs, plus one amendment the early cells forced.

## `frontier_3547.py`

The shared `agg/trajectories.csv` is final-only, and every question in this study is about the same cell at **two horizons**, so this reads the raw per-cell trajectories through `_cells_io.load_arm` — which is also what gets the base-row filter right (only ~1 in 34 rows is the production variant).

It reports realised harvest as a **first-class column** (#3547 asks for this explicitly), clicks-to-target and AUC at both horizons, the H3 t=100 replication anchor, the falsification arm, the H2 guardrail split, and H1 as a **difference-in-differences** rather than an argmin over a plateau #3319 already proved unresolvable.

## The amendment

Made off the first ~40% of cells and **before reading any verdict** — a limitation found after the answer is worth much less than one found before it.

The rebuild did what it was sized to do for the arms it was sized against:

| arm | k | #3319 harvest | **here** | cells >80% |
|---|---:|---:|---:|---:|
| `acq_m3` | -3 | 82.0% | **18.7%** | 0.0% |
| `acq_m4` | -4 | 85.3% | **34.9%** | 0.0% |
| `acq_m5` | -5 | — | 55.3% | 0.0% |
| `acq_m6` | -6 | — | 60.7% | 0.0% |

But `-5` and `-6` out-harvest what 900 positives was sized for and land above this plan's own 50% bar. The two arms added to probe *deeper* than the ship are themselves mildly compressed.

So the DiD's primary contrast becomes **`-4` vs `-3`**, the one pairing where neither side is compressed, and the verdict is taken on the uncompressed contrasts. That asymmetry is principled rather than convenient: compression caps a deep arm's late gains, biasing its DiD toward "no move" or "shallower" — one-sided, and in H1's *falsification* direction. **A compressed contrast can corroborate a "deeper" finding, can never produce one, and is never the evidence for "shallower".** The smoke test shows exactly that shape: `-5` and `-6` read "shallower" and are correctly excluded; the clean `-4`/`-3` contrast reads "no move".

## What it says about the sizing rule

Worth more than the caveat. 900 was chosen as the deepest value all twelve classes support — a **supply** bound — and checked against `preflight.sh` check 16b, a **horizon** bound. Neither is an **aggression** bound, and aggression is what actually sets harvest. A future deep grid should size from the harvest of its *deepest* arm, not its shipped one.

The compression here is also far milder than #3319's (60% vs 85%, and no cell anywhere above 80% against 29%), so it is a caveat on the deepest two arms rather than the study-killing confound it was there.

`tests_lib/meta`: 560 passed. Calibration index: 129 scripts, all classified.

Refs #3547, #3319, #2910, #2790.

https://claude.ai/code/session_01GidUmdxyBtfNeKKQbuhnbj


samggreenberg:
## The control landed — H2 is settled, and the study is complete

This push adds the H2 control's result, the report, and the five figures.
**1728 cells across both waves, 0 failures.**

### What the control proves

H2 was a cross-study comparison, and **two** things differed between #3319's
deep wave and this one: the pile *and* 79 commits of dev, including #3414, which
touched the very cost the inclusion knob prices. So the control held the code
fixed and moved only the pile — `vg_scale_any` at 400 clicks, on this commit:

| | harvest | spikes @100 | spikes @400 | first ≤100 | first >100 | median first-spike t |
|---|---:|---:|---:|---:|---:|---:|
| control `acq_m3` | **82%** | 0.0% | **5.7%** | **0** | **11** | **258** |
| control `prod` | 15% | 0.5% | 0.5% | 1 | 0 | 65 |
| deep `acq_m3` | 19% | 1.0% | 1.0% | 2 | 0 | 46 |

It reproduces #3319's number exactly, at #3319's harvest, on a commit 79 changes
later — and the spikes are genuinely *late*. On the deep pile, **zero of 1344
cells spike after t=100**. Dev drift is ruled out; **the hazard tracks exhaustion,
not aggression**, and #3319's deep guardrail is an artefact of its own ceiling.

Also worth flagging: within the control, the arm that is *not* exhausted (`prod`,
15% harvest) shows no late spikes either — an internal control on the same cells.

### Tooling, and one thing that was missing

Three of the analyzers were hardcoded to a single study path (`harvest_3547.py`
still to **#3319's** pile), so they all now take `--base`/`--arms`:

* `frontier_3547.py --csv` writes five tidy CSVs beside the markdown, so the
  figures and the tables read the same numbers. **Verified output-neutral** —
  `GENERATED_TABLES_3547.md` regenerates byte-identical apart from a trailing
  newline.
* `spike_timing_3547.py` reproduced its committed output exactly after being
  parametrised, before it was pointed at the control.
* `spike_examples_3547.py` (new) prints the literal trajectory rows around a
  cell's first spike. That is where the evidence came from that `n_good` sits
  **frozen at 88** through an exhaustion spike while the threshold oscillates.
* `figures_3547.py` (new) draws the five report figures.

`frontier_3547.py` has carried a comment since it was written saying the
deep-spike rule is "kept in sync by
`test_frontier_3547_spike_rule_matches_analyze_spikes`". **That test did not
exist**, and this branch was about to add a fourth restatement of the constants.
It exists now (`tests/core/test_experiment_spike_rule.py`), discovers copies by
scanning rather than by list so a new one is covered when it is written, and was
mutation-checked against a drifted threshold.

### Follow-up filed

**#3611** — size a deep grid from its **deepest** arm. 900 positives per class
was a *supply* bound checked against a *horizon* bound; neither bounds
aggression, which is what left `-5` and `-6` over the compression bar and cost
this study two of its three deep contrasts.

samggreenberg:
### Suite: green

`./run-tests.sh` on the GRID (job **614201**, `c0b8b5bec`): **10370 passed, 0
failed, 0 errors**, 56 skipped. pyright, pip-audit and vulture all OK. Frontend
gates skipped — no `node_modules` in this worktree, and this branch is Python and
docs only.

Worth recording because the first run looked otherwise. Job 614109 reported **0
failed but 2 errors**, both at *setup* of
`tests/datasets/test_parallel_loading.py::TestLoadingGates` ("a prior test leaked
a thread that is still holding the gate"). That is a **pre-existing flake, not
this branch** — now filed as **#3613**:

| run | tree | result |
|---|---|---|
| 614109 | this branch | 10368 passed, **2 errors** |
| 614111 | clean dev at `6a0168f7e` | 10364 passed, 0 errors |
| 614201 | **this branch, same commit, re-run** | **10370 passed, 0 errors** |

The two errored tests are exactly the difference between the two runs of the
*same commit* (10368 + 2 = 10370). The cause is that `_download_gate` and
`_embed_gate` are module-level globals shared per worker process, and none of the
~8 test files touching them is pinned to an `xdist_group` — so under
`-n auto --dist loadgroup` the pairing is decided by timing. A clean-dev baseline
alone would *not* have settled this; only re-running the same commit did.

One other gate caught a real omission: `check-calibration-index.py` blocks the
whole suite when a script in `scripts/experiments/calibration/` has no README
row. Fixed in `c0b8b5bec`.

Issue #3637, opened 2026-09-04
Folding an alias can un-band an image the class already saw — 248 across C, and `clock` nets −16


samggreenberg:
`pilebuild.loaders.vg_scale.canonicalise` merges an alias spelling's boxes into the class's, and `band_for` then reads the band off the union of all of them. `vg_scale`'s scatter guard rejects a union more than `BAND_MAX_INFLATION` (1.5) times the largest single box, on the correct grounds that such a union describes a scatter rather than an object.

The interaction was not measured until #3618 populated the alias table. Scored on the 56,579 non-COCO VG images (`name_coverage.py`, "BAND EFFECT"), of the **11,857** images the twelve classes currently band:

| | images |
|---|---|
| band unchanged | 11,586 |
| band **moved** | 23 (0.2%) |
| **left every band** on the scatter guard | **248 (2.1%)** |
| new banded positives from repaired images | +716 |

The 23 moves are the #3616 hazard and are negligible. The 248 are the question here. They are images the class **already saw and already banded**, which folding a second spelling turns into `SCATTERED` — so they leave the bands *and* the shared negative pool, becoming neither.

Per class the net is positive everywhere except one: **`clock` gains 18 banded positives and loses 34, net −16.** `kite` is +8. `bird` is +115 and `book` +76, so the aggregate ledger is **+468** and the fold is clearly worth making — but a class can be made worse by it, and nothing in the build says so.

## The question

Is a scattered fold the right outcome? Two readings, and they disagree:

- **Keep it scattered.** An image with a `clock` and a second `clocks` box at a different size was never a sound single-band positive; the guard is doing its job on information the builder previously did not have. This is the status quo and is defensible.
- **Keep the class's own band.** The class name's box is the evidence the cell was designated on. An alias box that is merely *elsewhere in the frame* adds a positive the class did not need and costs one it had. Under this reading `band_for` should be asked of the class's own boxes first, and the fold should only *add* boxes where the class had none.

The second reading also interacts with `designate_cells`: an image a human has already reviewed and that leaves every band is a reviewed positive lost to a config edit, which is the failure #3616 and #3614 both circled.

Whichever is chosen, the build should **report** it: `canonicalise` already returns `{class: boxes folded in}`, and the count of images the fold un-bands belongs beside it, per class, in the same log line — the same argument as "a merge that folds nothing is worth seeing".

[...]

Issue #3636, opened 2026-09-04
Adjudicate head-noun compounds as one hypothesis — 76 VG names are individually unmeasurable but pool to 74%


samggreenberg:
#3618 adjudicates each candidate VG name on its own, against a floor of five images where that name is the class's only evidence. **76 of 626 candidates fall below the floor** and are recorded as `unmeasured` — neither acted on nor refuted.

They are not noise. Pooled, they carry **312 non-COCO images** at **58%** precision; setting aside `stop sign`'s dead `sign` family (0 of 37, 90 images), the other eleven classes' 222 images pool to **74%**. That is roughly **160 more images of real repair**, sitting behind names English would call obvious: `yacht`, `rowboat`, `ferry`, `cruise ship`, `beach umbrella`, `green umbrella`, `white dog`, `black kite`, `grandfather clock`, `pigeons`, `dove`.

The floor is right for a name judged alone — 2 of 2 is not a rate. But most of these names are not independent hypotheses. `blue umbrella`, `red umbrella`, `green umbrella`, `orange umbrella` and `yellow umbrella` are one hypothesis five times over: *a colour modifier in front of the class name does not change what the name denotes*. That hypothesis is testable at the sample size of the whole family, not one colour at a time.

## What to do

Add a pooled mode to `scripts/experiments/pile/name_evidence.py`:

- group a class's head-noun compounds by construction — colour modifier, `an X` / `X.` / `X's` typing, size modifier (`small X`, `large X`) — rather than by individual name;
- adjudicate the **group**, then apply the verdict to each member;
- keep the individual measurement as the veto: a member with enough images of its own that disagrees with its group (`hot dog` inside `* dog`, `gravy boat` inside `* boat`) must fall out, or the pooling reintroduces exactly the trap #3618's head-noun section documents.

The grouping is where the judgment is, so it should be spelled out in the config next to the tables it fills, not inferred by regex at run time.

Worth roughly 160 repaired images across the eleven classes that have any — about a fifth of what the shipped tables buy (860). Cheap to measure: one pass over `objects.json`, ~2 minutes on a CPU node.

Follows from #3618. See `docs/experiments/2026-09-04-vg-name-coverage/REPORT.md`, "What is left on the table".

samggreenberg:
Addressed in #3664.

[...]

Issue #3635, opened 2026-09-04
`stop sign` cannot be repaired by any VG name — 18.7% self-match, and its 15k-image `sign` family scores 7.9%


samggreenberg:
`stop sign` is the worst-served class in *C* and #3618 could not repair it with a name.

Measured over the 51,411-image VG∩COCO overlap and the 56,579 VG images COCO does not annotate (`docs/experiments/2026-09-04-vg-name-coverage/REPORT.md`):

- **496 VG images** carry the name `stop sign` — the smallest supply of the twelve by a factor of three.
- **18.7% self-match**: of COCO's 1,016 `stop sign` boxes on the overlap, only 190 carry a VG box named `stop sign`. Lowest in *C*.
- **46.6% carry a VG box named `sign`** — the largest fold-in column anywhere in *C*.

And `sign` is not usable. On the images where `sign` is the class's only evidence, COCO finds a stop sign **7.9%** of the time (Wilson lower bound 0.07), so listing it in `pile_config.SCALE_VG_AMBIGUOUS` would withhold **12.7 images from the shared negative pool per contaminated negative removed** — and the pool is shared, so that cost lands on all twelve classes, not just this one. The rest of the family is no better: `signs` 9.0%, `street sign` 8.0%, `traffic sign` 6.4%, `red sign` 23% (lower bound 0.14). The 37 adjudicable images behind every remaining `* sign` name pool to **0 of 37**.

What #3618 could take was `stop` (70% precision, 6% box agreement — the lettering) and `octagon` (80%, only 15 boxes). Both are ambiguous-table entries: they withhold 157 images and produce **zero** positives. `stop sign` is the one class in *C* that gains nothing from the name tables.

## What to do

The missing positives are real — COCO says so on the half where it can — and no VG name reaches them. Options, roughly in order of cost:

1. **A human pass**, the same shape as #3588's slates: rank the `sign` family by a `stop sign` text query, box what is a stop sign. The candidates are already enumerated (`vg_name_families.py` prints 155 names on 19,148 images; 8,070 of those images are off-COCO).
2. **Restrict the class to the COCO half**, where `anchor_to_coco` already makes it exhaustive, and accept that its pool is smaller and differently distributed from the other eleven — which breaks the paired-cell construction.
3. **Drop `stop sign` from `SCALE_CLASSES`** at the next rebuild. It was chosen for supply that measurement now says is half illusory.

[...]

Issue #3616, opened 2026-09-03
Re-boxing a positive to a better example silently moves it between bands


samggreenberg:
Sam, after finishing the `bowl` slate: *"There were a few times where I changed your box,
not because your box was 'wrong', but because there was a different, more-prominent example
elsewhere in the image."*

That is a reasonable thing for a reviewer to do, and it quietly rebalances the bands.

## What it does

An image in the `positive_boxed` stratum was drawn into a specific `class@band` cell
**because of the box it arrived with**. `verdicts_to_corrections.py` treats a redrawn box as
`human_rebox` — *"reviewer redrew it: the box, hence the band, changes"* — so retargeting
the box to a different object moves the image to whatever band the new box implies, and it
leaves the cell it was sampled to fill.

Measured over the four classes reviewed so far (band edges `small ≤ 1/196 < medium ≤ 1/12 <
large ≤ 0.80`):

| class | redrawn | changed band | |
|---|---:|---:|---|
| bench | 1 | 1 | 2401443 small → medium |
| sink | 5 | 2 | 2333898, 2403014 large → medium |
| bowl | 7 | 3 | 2348717, 2358315 medium → large; 2378456 small → medium |
| cell phone | 0 | 0 | |
| **total** | **13** | **6** | |

**Nearly half.** Two of the six leave the `small` band, which is the binding constraint on
class supply everywhere in this project (#3603) and the band the review protocol already
erodes (#3156 rejected 43% of small-band positives as a function of object size).

## The cause is partly the guide

The protocol said *"when several objects qualify, box the single biggest one."* That is
right when the reviewer is **finding** an object in a bare image, and wrong when they are
**adjudicating** one that arrived boxed — two different questions that the rule did not
distinguish. Fixed in the guide: on a pre-boxed positive, judge the object in the box and
redraw only to correct the extent of the *same* object.

## Still worth deciding in code

A guide sentence is a weak guarantee. Options, roughly in order of how much they cost:

1. **Report it.** Have `verdicts_to_corrections.py` count `human_rebox` entries whose band
   differs from the sampled cell, so the drift is visible instead of silent.
2. **Refuse it.** Treat a rebox that changes band as a rejection of the sampled object
   plus a *new* boundary-style discovery, which is what it actually is.
3. **Re-draw the cell.** Accept the move and top the vacated cell back up, which means the
   slate is no longer a fixed sample.

(1) is cheap and should happen regardless. (2) is the honest modelling but changes what a
[...]

PR #3638, opened 2026-09-04
Measure every vg_scale class's VG names, and fill both tables (#3618)

### LINKED ISSUE #3618: Measure VG-name coverage for the other eleven `vg_scale` classes before the next rebuild
**Difficulty: low-medium** — one measurement run over data this repo's sessions cannot reach, then a config edit and a reading.

**Recommended Claude model: Sonnet 5.** The run is a single script invocation and the edit is a table; the judgment is reading a fold-in column and deciding alias vs. ambiguous vs. neither, which is bounded and has a worked example (`bicycle`) to calibrate against. Step up to Opus 4.8 only if a class turns out to need a merge, since `canonicalise` folds boxes into the positives a rebuild is designated from.

## Why

#3605 found that `vg_scale` built `bicycle` from the VG name `bicycle` alone while `bike` carried 638 of COCO's 3,683 `bicycle` boxes on the overlap. That is not a supply loss: on the ~52% of VG that COCO does not annotate, VG's silence is the only evidence of absence, so an instance under an unlisted spelling becomes a **negative for its own class**.

#3617 fixes the mechanism and declares `bicycle`. It measures nothing else. The other eleven classes were never checked, and `pile_config.SCALE_VG_NAMES_AUDITED` now says so out loud — a `vg_scale` build prints:

```
VG-NAME COVERAGE UNMEASURED for 11 of 12 classes: clock, bird, boat, umbrella, book, dog, backpack, knife, bus, stop sign
```

**Before the next rebuild** is the moment this is cheap. Afterwards it costs a re-embed, and #3604 already has one rebuild pending a decision.

## What to do

The measurement exists — `coco_folds.py`, added by #3606 (merge or rebase onto it first):

```bash
source scripts/experiments/pile/pile_env.sh
python coco_folds.py --out folds-c.json      # defaults to SCALE_CLASSES
```

Read the **fold-in** column per class: the VG names landing on that class's COCO boxes over the ~51k VG∩COCO overlap. Any name reaching a material share is a spelling the builder is currently dropping.

For each such name, decide which of three it is, and confirm the alias reading with `scan_name_overlap.py` (box IoU both ways) rather than on the fold-in count alone:

| finding | where it goes | effect |
|---|---|---|
| same object, other spelling (high IoU both ways) | `pile_config.SCALE_VG_NAMES` | `canonicalise` folds its boxes onto the class |
| may be the class, may be something else | `pile_config.SCALE_VG_AMBIGUOUS` | `lift_ambiguous` withholds the image from the bands and the negative pool |
| co-occurring but distinct | nowhere | — |

Then add every class you measured to `SCALE_VG_NAMES_AUDITED`, whatever the verdict — an audit that found nothing is exactly the result the flag exists to record, and is what distinguishes "no spelling exists" from "nobody looked."

## Notes

- **The known likely hit is `book`.** #3606 already ran `coco_folds.py` against it: `magazine` (79) and `magazines` (30). That is a *definition* question (COCO has no magazine class, so COCO's annotators put magazines in `book`) rather than a spelling one, and it already cost that class 49 verdicts — so decide deliberately whether it belongs in either table or is purely an annotation-guide matter.
- **`bus` has a known false lead.** String containment once matched 80 images annotated `bush` (`docs/experiments/2026-08-12-overview-bench/REPORT.md`). Both tables are measured for this reason; do not populate either from a name that merely looks related.
- **Plural forms count.** `magazine`/`magazines` split in the data, so a plural is a spelling like any other.
- This changes what a rebuilt `vg_scale` contains. It does not change the published pickle, whose `bicycle` caveat is recorded in `docs/experiments/2026-08-25-vg-scale/DATASHEET.md`.

Follows from #3605; mechanism landed in #3617.

samggreenberg:
Closes #3618.

#3617 fixed the mechanism by which a `vg_scale` class built from one VG spelling turns its own instances into negatives, declared `bicycle`, and recorded that the other eleven had never been looked at. This measures all twelve and fills both tables.

`SCALE_CLASSES` is untouched and **nothing is rebuilt** — what changes is what a rebuild would contain, which is why #3618 wanted it before the next one (#3604).

## The measurement #3618 specifies cannot make the decision on its own

Fold-in is a **box** test, and it is conditioned the wrong way round:

| name | class | fold-in | truth |
|---|---|---|---|
| `sign` | `stop sign` | **46.6%** — the largest fold-in column anywhere in *C* | a VG `sign` box is a stop sign **7.9%** of the time |
| `grandfather clock` | `clock` | 0 of 6 | every one of those images holds a clock — the VG box is the cabinet, COCO's is the dial |

A high fold-in share says *stop signs are usually annotated `sign`*, which is a fact about stop signs, not signs. So `name_evidence.py` adds the image-level test the negative pool actually needs: over the VG∩COCO overlap, take the images carrying a name and **not** the class name — the state that becomes a false negative on the other half — and ask COCO whether the class is present.

**`1 / precision` is the price**: images withheld from the *shared* pool per contaminated negative removed. `sign` costs 12.7, `bike` 2.1, `books` 1.1. The cut is 3, on the **Wilson lower bound**, so a name measured on five images cannot outrank one measured on two thousand. Box agreement then decides which table, because folding claims *this box is the object* and a band is a claim about one object's size (#3616).

## What it buys

32 spellings fold, 50 are withheld, 468 refuted, 76 unmeasurable. On the 56,579 VG images COCO does not annotate:

- **860 images repaired** — they were negatives for their own class. `bird` +18% on what it could already see, `book` +12%, `boat` +11%.
- **2,664 withheld** (4.7%; 1,198 of them `bike`, already shipped by #3617).
- Overlap coverage against COCO rises for eleven of twelve: `bird` 32.1 → 39.1%, `boat` 30.9 → 36.2%, `dog` 74.8 → 79.2%.

`stop sign` gains nothing, and that is the study's sharpest result: its 15,042-image `sign` family is refuted at 7.9%, so 157 images leave its pool and not one becomes a positive (**#3635**).

## Three method results worth keeping

- **The box-overlap confirmation #3618 asks for cannot see a spelling split.** 846 of 1,740 candidate pairs never co-occur, because an annotator who writes `back pack` does not also write `backpack`; `backpack`/`backpacks` scores **0.000 both ways**. Where it works is where two names genuinely sit on one box — `clock`/`clock face`, 0.562/0.701 — and refuting a lookalike, which is how `bus` survived `bush`.
- **There is no free fix in the reader.** All **2,516,939** VG objects carry a `names` list of length one, and zero of the 18,897 objects named by a table entry carry the class name further down it.
- **A head-noun family is a list of names to measure, never a list to fold.** The largest member of `dog`'s is **`hot dog`** (405 images), which scores **0 of 181**.

## Also

`name_coverage.py` prices a table before it ships, including the **band ledger** a fold pays out of: merging boxes pushes 248 images past the scatter guard and out of every band, and `clock` nets **−16** (**#3637**). Run with no `--propose` it scores what `pile_config` actually ships — which is how this PR's tables were verified to reproduce the proposal they came from, exactly.

Report, figures and the JSONs they are drawn from: `docs/experiments/2026-09-04-vg-name-coverage/`.

Follow-ups: **#3635** (`stop sign` needs a human pass, not a name), **#3636** (pool the 76 unmeasurable compounds — ~200 more images), **#3637** (the scatter-guard interaction).

https://claude.ai/code/session_01WtEFFR8kA2y2aHAQezpxcH


samggreenberg:
Full GRID suite green on `29a4ce841` (job 616222): **10,603 passed**, 6 skipped, 1 xfailed; pyright, pip-audit, vulture, npm audit, check-docs, codespell, deptry and the Vitest frontend suite all OK.

The first run on `eb592c9ba` failed 4 tests in `tests_lib/meta/test_pile_class_rules.py` (`'str' object has no attribute 'name'`). That was **not this branch**: `af42d4b15`, the dev commit it was cut from, carried two module-level `SCALE_CLASS_RULES` assignments and the later `dict[str, str]` one shadowed the `dict[str, ClassRule]` one. `d06935623` on dev fixes it; this branch now merges dev, and the tables it adds were re-verified by name after the merge (32 alias / 50 ambiguous / `SCALE_VG_NAMES_AUDITED` == `SCALE_CLASSES`).

Issue #3670, opened 2026-09-06
Drop vg_scale prevalence to 1%: the negatives exist and cost one embedding run


samggreenberg:
Drop vg_scale prevalence to 1%: the negatives exist and cost one embedding run

## The ask

From the reviewer, on finishing the #3588 negative pass:

> I'd like to have more negatives in the set, if possible. Instead of 2.5%
> prevalence, I'd like 1%.

At `SCALE_N_POS = 100` per cell, 1% needs **9,900 negatives** against the
current `SCALE_N_NEG = 3,900`.

## The supply is there, and most of it is provable

| | images |
|---|---|
| Visual Genome, reachable | 108,245 |
| in the pile today | 7,747 (4,200 pool + 3,547 positives) |
| **outside the pile** | **100,498** |
| ...carrying no VG name for any of the 25 | **69,266** |
| ...**and COCO confirms none of the 25 present** | **18,986** |

So the provable negative pool could be **4,200 + 18,986 = 23,186** — more than
twice what 1% requires, and enough for 0.43% if that were ever wanted. No human
labelling is involved: COCO annotates all eighty of its classes on any image it
annotates, so "holds none of the 25" is a fact on that subset.

## Recommended composition: match the positives' provenance

Taking *only* COCO-confirmed negatives would make the negatives 100%
COCO-anchored while the positives are 57% (#3668). VG's images come from COCO
**and** YFCC100M, so provenance is a real visual variable and a detector could
learn it instead of content.

Matching the positives' 57/43 split at 9,900 negatives:

| | need | available | add |
|---|---|---|---|
| COCO-anchored | 5,643 | 20,831 | **+3,798** |
| non-COCO | 4,257 | 52,635 | **+1,902** |

**5,700 new images to embed.** On the app's own GPU node that is minutes, not
the ~12 CPU-minutes-per-200 that #3669 documents.

## What it costs in label quality: nothing new

The non-COCO half carries VG-silence error, which #3588 has now measured
directly: **0.5–2.5% per class** on the shared pool (16.5% for "holds any of the
25", but only the per-class figure bears on a per-class eval). Adding non-COCO
negatives in the same proportion keeps that rate identical — proportionally the
pool is no worse than today, just larger.

Sequencing note: this should land **after or with #3667**, since that changes
which images are eligible as negatives at all.

---

**Label corrected 2026-09-06: this is an `experiment`.** It was filed with a
[...]

Issue #3588, opened 2026-09-03
Expand `vg_scale`'s class list: the 12 classes now confound size with context


samggreenberg:
**Difficulty: medium-high** — a dataset rebuild plus a human correction pass, on a builder whose expensive bugs are documented.

**Recommended Claude model: Opus 4.8.** The build passes (`apply_corrections`, `designate_cells`) are exactly where this dataset's costly bugs have lived, and a wrong-but-plausible change there is silent. The first step alone — re-run the shortlist and read the table — is Sonnet 5 shaped.

## Why

Reading the #3156 map per-class rather than per-band (`siglip` / whole image, cost at 150 clicks, 20 seeds) turns up something the band means hide:

- **Within a band, which class you picked matters more than which band you are in.** Every `small` positive is under 1/196 of the frame *by construction*, yet the twelve classes span **9.4×** there (`kite` 0.067 → `backpack` 0.627). The pooled band effect the study was built to measure is **4×** (0.43 → 0.11). The confound the dataset was designed to remove is smaller than the one it left in.
- **Easy-when-large does not predict easy-when-small.** Over the twelve classes, ρ(small cost, large cost) = **−0.02**. `bird` is the cheapest cell in the entire study at large (0.026) and 31st of 36 at small (0.520).
- The flattest classes are the two with near-exclusive scenes — `kite` (small/large ratio 1.8×) and `boat` (3.2×) — and the most expensive small cells are objects whose surroundings *are* the negative pool: `backpack` 0.627, `stop sign` 0.588, `clock` 0.553.

That last bullet is a reading of the ordering, not a measurement — this grid varies size and class together and cannot separate them. The ablation that would is a separate issue (see below). **This** issue is the cheaper half: the class list currently samples the context axis by accident (2 of 12 classes are scene-exclusive), so it should sample it on purpose.

## The gate any addition must clear

From `scripts/experiments/pile/pile_config.py` and `shortlist_scale_classes.py`, a candidate must be simultaneously:

1. **≥100 images in all three bands** under `--compact` (the binding constraint — ranked on the *minimum* per-band count).
2. **A COCO-2017 class.** This is what makes the correction pass affordable: COCO val2017 is exhaustively annotated, so VG's miss rate and the reviewer's own accuracy are scored for free on the 48% of images that are COCO-sourced. Dropping this is what turns a rebuild into a full manual annotation job.
[...]

PR #3672, opened 2026-09-06
Let a class be scored against images holding a different class (#3667)

### LINKED ISSUE #3667: vg_scale never scores a class against images holding a different class
vg_scale never scores a class against images holding a *different* class

## The rule

`pilebuild/loaders/vg_scale.py:465`

```python
"evaluable_categories": cats if cats else (list(cells) if iid in neg_set else []),
```

An image that is a positive for **any** class is scorable **only in its own
cells**. So an image holding a `book` and no `bus` is neither a bus positive nor
a bus negative — it is dropped from the bus evaluation entirely.

Every class's negatives are therefore the *same* 4,200 shared-pool images, and a
shared-pool image holds **zero** of the twelve by construction.

## What that costs

| | |
|---|---|
| pile | 7,747 images |
| carry a cell (positives) | 3,547 |
| shared pool (the only negatives) | 4,200 |
| **dropped from each class's evaluation for holding a different class** | **3,247 — 41.9%** |

Identical for all twelve classes, because cells are designated at 300 positives
each.

**The consequence is a distribution shift between the two sides of every
evaluation.** Positives are images containing a labelled object; negatives are
images containing *none* of twelve common object classes. A detector can score
well on that contrast by learning "is this a scene with stuff in it" rather than
"is there a bus". Nothing in the benchmark asks whether a bus detector fires on
a photo of a dog, because such an image is never scored for bus.

## The stated design goal does not require this

The loader's docstring gives the reason for the shared pool:

> Every cell therefore has identical prevalence and identical negatives, so a
> small-vs-large difference is a paired contrast on one class rather than two
> datasets of different difficulty.

That goal is about **bands within a class**. Adding non-bus positives to bus's
negatives keeps `bus@small` and `bus@large` scored against identical negatives,
so the paired contrast survives intact. The exclusion is stronger than the
property it was introduced to protect.

The docstring separately justifies excluding an image that holds *c at another
size* — "scoring it as a negative would penalise a detector for finding a real
bus". That reason is sound and specific to the same class. The code applies it
to every other class too, where it does not hold.

## The fix is free for about half of it

Label risk is the real reason to be careful: VG's silence about `bus` on a
`book` image is weak evidence. But **on the COCO-anchored half there is no
risk** — COCO annotates all eighty of its classes on any image it annotates at
all, so "does this hold a bus?" is answered exactly.

| class | blind | blind **with an exact COCO answer** | negatives now | could be | gain |
|---|---|---|---|---|---|
| clock | 3,247 | 1,880 | 4,200 | 6,080 | +44.8% |
| bus | 3,247 | 1,867 | 4,200 | 6,067 | +44.5% |
| book | 3,247 | 1,848 | 4,200 | 6,048 | +44.0% |
| stop sign | 3,247 | 1,791 | 4,200 | 5,991 | +42.6% |

…and so on; every class gains 42–45%.

**About 1,850 hard negatives per class, at zero labelling cost and zero label
risk.** And they are exactly the cluttered, realistic negatives the benchmark
currently has none of.

Note the direction of the annotation quality, which is the opposite of the
intuition: **57.0%** of the positives are COCO-exhaustive against **43.9%** of
the shared pool. The images being excluded are *better* annotated than the
negatives being kept.

## Trade to decide

Adding per-class negatives means the twelve classes no longer share one negative
set, so cross-class comparisons change while the within-class band contrast —
the one the construction is *for* — is preserved exactly. Worth confirming that
is acceptable before building it.

Related: #3666 (pool error for the shipped twelve has never been measured) and
\#3588, whose thirteen-class review found per-class pool error of 0.0–7.1% and
established that a missing VG name is not an absent object.

---

**Label corrected 2026-09-06: this is an `experiment`.** It was filed with a
`not-an-experiment` marker, which was wrong — the code fix is merged, but the issue closes only when the pile is rebuilt — \`build_pile.py --datasets vg_scale --force\`, which re-embeds 7,747 images on a GPU node. The marker named the rebuild and then denied it.

The marker was written to satisfy the label hook rather than to apply its test.
The hook asks whether the issue can be closed without a run; six issues in a row
got the marker, and five of them cannot.

samggreenberg:
Closes #3667.

## The rule, and where it went wrong

`vg_scale` designates an image as a positive for its own cells, a negative only when it holds nothing in *C*, and **excluded otherwise**.

That exclusion is right for the **same class at another size** — scoring a large-bus image as a small-bus negative penalises a detector for finding a real bus, which is exactly what #3156 exists to prevent. But it was applied to **every other class too**, where the reason does not hold. An image with a book and no bus was neither a bus positive nor a bus negative; it simply left the bus evaluation.

Two costs:

- **41.9% of the pile** was dropped from every class's evaluation.
- A distribution shift across every contrast: positives are images containing a labelled object, negatives contain none of twelve common objects. A detector could score well by learning *"is this a scene with stuff in it"* rather than *"is there a bus"*.

## The fix

`_evaluable()` applies the exclusion **per class**. A class the image does not hold contributes all of its cells and the image scores there as the negative it is. A class it *does* hold still contributes only the cells it was drawn for.

**Gated on `labels_exhaustive`.** COCO annotates all eighty of its classes on any image it annotates, so "holds no bus" is a fact there. Off COCO it is VG's silence, which #3588 measured wrong **0.5–2.5%** of the time per class — importing that into the negatives is a separate decision, and `SCALE_CROSS_CLASS_NEGATIVES` turns the whole thing off rather than letting the two halves be treated as alike.

## What it buys

Priced from the shipped cell **without rebuilding it**, since several studies are reading that file right now:

| | before | after | |
|---|---|---|---|
| evaluable images per cell | 4,300 | ~6,150 | **+43.0%** mean over 36 cells |
| negatives per cell | 3,900 | 5,780 | |
| **prevalence** | **2.50%** | **1.70%** | |

That last row is most of the way to the 1% requested in #3670, at no labelling cost and with no new images.

## Tests

11 new tests pin the asymmetry that matters:

- a bus image is **still barred** from bus's other bands (the #3156 guarantee)
- a bus image is **now a perfectly good book negative** (the #3667 fix)
- an image holding both bus and book is barred from both, and still usable for dog
- **off-COCO images are left alone** — the gate
- the knob turns it all off

## Not deployed

The shipped `vg_scale__siglip.pkl` still carries the old field. A `build_pile.py --datasets vg_scale --force` rebuild is what makes this real, and it should be **coordinated** — the file is live for other work, and rebuilding it under a running study is the kind of thing that produces an unexplainable result three days later.

https://claude.ai/code/session_01FAnc7gMSFh43SCLGBM3zCv


samggreenberg:
Full suite green on `95a3458a9`: **10,758 passed** (11 new), 56 skipped, 2 xfailed. pyright, pip-audit and vulture all OK. Job 618470.

The first run failed pyright only — integer box literals in the new test file, since `list` is invariant so `list[list[int]]` is not `list[list[float]]`. Fixed in `95a3458a9`; no behaviour change.

Issue #3698, opened 2026-09-07
The vg_scale pile rebuild is owed for #3670 and #3588 — build it once, and check it happened


samggreenberg:
**Recommended Claude model: Sonnet 5.** The build is one launch and ~16 minutes;
what needs care is the order and the verification afterwards.

## What is owed

Two merged-or-merging changes to `vg_scale`'s construction have no cells behind
them:

- **#3670** — `SCALE_N_NEG` 3,900 → 9,900, all COCO-scored (`SCALE_NEG_COMPOSITION`
  = `provable`), plus a per-media `coco_scored` stamp.
- **#3588** — the class list goes from twelve to twenty-five.

They were deliberately **not** rebuilt separately. #3670's build ran on
2026-09-06 at 19:15 and a parallel 25-class rebuild overwrote its cells 37
minutes later; doing that twice more is a waste and, worse, leaves a window in
which each study's report describes cells nobody has.

## Until it happens, the pile disagrees with the code

`--verify` now says so rather than passing quietly (#3670 added both checks):

- the designated pool must have `SCALE_N_NEG` images, and the failure quotes the
  prevalence the cell actually has;
- under `provable`, every designated negative must carry `coco_scored` — and a
  cell built before that stamp existed is told to rebuild rather than passed on
  the weaker `labels_exhaustive`.

So the hazard is visible. It is still a hazard: any study that runs against the
live pile in the meantime is measuring 2.5% prevalence while `SCALE_PREVALENCE`
says 1%.

## Doing it

```bash
bash scripts/experiments/pile/launch_pile.sh --datasets vg_scale --force   # pulls vg_scale_any in
python scripts/experiments/pile/build_pile.py --verify
python scripts/experiments/pile/check_review_coverage.py
```

Four things to get right, each of which has bitten before:

- **Archive the current cells first.** `/expscratch/sgreenberg/archive/` — and
  note the glob `vg_scale__*` does **not** match `vg_scale_deep__*` (#3667 lost
  the deep cell from its archive that way).
- **`VTS_SCALE_ROSTER` decides what the review still covers.** The live roster is
  whatever built last; start from the roster the change starts from.
- **A rebuild is never only a relabel** — it runs against `dev` and picks up
  every `pile_config` ruling merged since the cell was built (#3667: five of
  them, 46 of 360 reviewed positives).
- **`vg_scale_deep` does not follow** #3670's pool size (#3690) and needs its own
  decision before it is rebuilt for #3588's classes.

[...]

Issue #3697, opened 2026-09-07
#3667's cross-class rule reads `exhaustive`, which a one-class review also sets


samggreenberg:
**Recommended Claude model: Opus 4.8.** The change is two lines; deciding
whether to make it means re-reading what #3667 admits and pricing a rebuild.

Split out of #3670, which fixed the same defect in one place and left it in the
other.

## The defect

`labels_exhaustive` is set by two different things:

- an image anchored to a COCO annotation — COCO answers for all eighty of its
  classes at once, so absence is a fact for every class in *C*;
- **any image a human looked at** — `apply_corrections` calls
  `exhaustive.add(iid)` for every verdict.

Those are not the same claim. A reviewer asked *"does this hold a `car`?"*
established a fact about `car` and nothing about `bus`.

`_evaluable` gates #3667's cross-class rule on `exhaustive`:

```python
if pc.SCALE_CROSS_CLASS_NEGATIVES and iid in exhaustive:
    held = set(labels.get(iid, {}))
    for c, owned in _cells_by_class(cells).items():
        if c not in held:
            out |= owned
```

So an image reviewed for one class becomes a scorable **negative for the other
eleven** on the strength of VG's silence — exactly the inference #3667's
`exhaustive` gate exists to refuse.

## How big it is

Measured on the shipped twelve (`negpool_coverage.py`, `over_promoted` block):

| | images |
|---|---:|
| COCO-scored | 51,448 |
| `exhaustive` | 51,645 |
| **promoted by a review alone** | **197** (0.4%) |
| …of which are designated positives | **64** — 1.8% of designated positives |

Small. Not zero, and it is 64 images serving as cross-class negatives for up to
eleven classes each.

## Why #3670 did not just fix it

#3670 needed the strict set for the negative pool, so it captures `coco_scored`
before corrections widen `exhaustive` and stratifies the draw on that, with a
per-media `coco_scored` stamp `--verify` checks. It deliberately did **not**
change `_evaluable`: that would change which images #3667 admits as negatives,
which changes every cell's evaluable pool and its realised prevalence, and is a
rebuild with its own before/after measurement — not a side effect of a
prevalence change.

## What to do

Decide whether `_evaluable` should read `coco_scored` instead. The stamp and the
set already exist. Price it the way #3667 was priced (`cross_class_negatives_effect.py`
reads the shipped cell before any GPU), and if it moves, fold it into the
rebuild that #3588 and #3670 are already waiting on rather than taking a third.

[...]

Issue #3693, opened 2026-09-06
launch_pile.sh built the pile from a checkout 1,420 commits behind dev (fixed default; needs a provenance stamp and a staleness gate)


samggreenberg:
**Recommended Claude model: Sonnet 5.** The fix is one line; the value is in the
check that stops it recurring.

## What happened

`launch_pile.sh` set the checkout its jobs import from a **fixed path**:

```sh
REPO="${VTS_REPO:-/exp/$USER/projects/vts-pile}"
```

So `bash launch_pile.sh vg_scale`, run from any other worktree, submits jobs that
build the pile from `/exp/$USER/projects/vts-pile` — **1,420 commits behind
`origin/dev`** as of 2026-09-06, predating `vg_scale` entirely (its `pile_config.py`
has no `VTS_SCALE_N_NEG`). Nothing in the launch output names the checkout.

Caught during #3670 only because that stale copy is old enough to lack
`build_pile.py --rebuildable`, which the launcher's own canary calls — so it
failed loudly. **Had the two been any closer in age it would have rebuilt every
`vg_scale` cell from code nobody was looking at and reported success**, exactly
the way #3269 measured a retired head from a worktree 321 commits behind.

## Fixed here

`REPO` now derives from the script's own location, `VTS_REPO` still overriding:

```sh
REPO="${VTS_REPO:-$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)}"
```

You launch the code you are looking at.

## What is still missing, and is the real ask

The launcher prints `submitted vg_scale -> job NNN` and never says **which
checkout at which commit**. Two things worth adding:

1. **Print the resolved `REPO` and its `git rev-parse --short HEAD`** in the
   launch banner, and stamp both into the cell's provenance. #3269's lesson was
   "record a resolved default as data, not just as a launcher argument", and the
   pile's provenance records the *machine* that built a cell but not the *code*.
2. **Refuse a `REPO` far behind `origin/dev`**, the way `preflight.sh` check 4
   fails a study worktree more than `PREFLIGHT_MAX_BEHIND` commits behind. The
   pile has no equivalent gate, and it is the artifact every study reads.

Worth checking whether any other launcher in `scripts/` carries the same fixed
default — `launch_scale.sh` and `launch_bands.sh` are the obvious candidates.

samggreenberg:
[...]

Issue #3690, opened 2026-09-06
Should vg_scale_deep follow #3670's prevalence change, or stay pinned at 11,700 negatives?


samggreenberg:
**Recommended Claude model: Sonnet 5** if the answer is "follow"; **Opus 5** if the
horizon comparison has to be re-read first.

#3670 takes `vg_scale`'s shared negative pool from 3,900 to 9,900, so a band-cell
sits at 1% prevalence instead of 2.5%. `SCALE_PREVALENCE` (the band-free number)
therefore moves 0.0714 → 0.0294.

`SCALE_DEEP_N_NEG` was **derived from `SCALE_PREVALENCE`**, so that change would
have taken `vg_scale_deep` from **11,700 negatives to 29,700** as a silent side
effect of a change to a different dataset — a 2.5× rebuild nobody asked for.

## What was done instead

Deep's prevalence is pinned to the pre-#3670 value via `SCALE_DEEP_PIN_N_NEG`
(3,900), and `SCALE_DEEP_N_NEG` is still *derived* from it — the "never write a
negative pool as a literal beside a positive count" rule is intact; what is pinned
is the `vg_scale` pool size deep's prevalence refers to.

The reason is the same one `SCALE_DEEP_N_POS`'s docstring already gives for
holding the class list fixed: **deep exists for one comparison** — the #3319/#3547
acquisition horizon — and moving its prevalence mid-stream would confound the
horizon axis with a prevalence axis.

## The question

Should `vg_scale_deep` follow?

**For following:** deep's stated job is to *reproduce* `vg_scale`'s prevalence
rather than re-derive it, and a family whose two members disagree about prevalence
is the thing `SCALE_PREVALENCE` was named to prevent. #3667 already moved both
cells together (4.99% / 5.09%) and that togetherness is what let the "only depth
changed" premise survive — a pin breaks it deliberately.

**Against following:** 29,700 negatives is a 2.5× rebuild of a dataset built for
one comparison, and every #3319/#3547 number would need re-reading against a new
prevalence. #3681 already found that a realised-prevalence move of half a bit
shifted k\*.

## What would settle it

Whether any live study reads `vg_scale` and `vg_scale_deep` in the same table. If
none does, the pin is free and can stay indefinitely. If one does, the pin is a
confound and deep should follow before that study reports.

Filed from the #3670 implementation; see `pile_config.SCALE_DEEP_PIN_N_NEG`.

samggreenberg:
[...]

PR #3699, opened 2026-09-07
1% prevalence from a provable negative pool (#3670)

### LINKED ISSUE #3670: Drop vg_scale prevalence to 1%: the negatives exist and cost one embedding run

samggreenberg:
Takes `vg_scale`'s shared negative pool from **3,900 to 9,900** so a band cell
sits at 1% designed prevalence instead of 2.5%, and draws every one of them from
the COCO-scored half of Visual Genome, where "holds none of *C*" is a fact
rather than a silence.

Report: [`docs/experiments/2026-09-06-provable-negatives-3670/REPORT.md`](docs/experiments/2026-09-06-provable-negatives-3670/REPORT.md).
Suite green on the GRID: **10,857 passed**, all gates.

## Three things a reviewer should know before the diff

**The realised prevalence is 0.85%, not 1%.** Since #3667 a cell also scores the
other eleven classes' COCO-scored positives — ~1,900 of them — so 9,900 shared
negatives land every cell at **0.847%** (0.844–0.856 over 36 cells, k\* −6.87).
A realised 1.00% would need ~8,000. This ships the 9,900 both #3670 and #3690
specify, which overshoots in the direction the ask pointed, but the number to
quote is 0.85%. See #3681.

**The composition is chosen on the spread, not the magnitude, and the trade as
first written down was wrong.** It said provenance shortcut 1.11× against
contamination 1.32×, reading as decisive. That 1.32 took the top of #3635's
*predicted* per-class range; #3666 has since **measured** the shipped twelve at
1.40% pooled, which puts contamination at **1.18× [1.09, 1.37]** — an interval
containing the shortcut's **1.1×**. On magnitude the two compositions are not
separable. What separates them is that the provenance shortcut is *uniform
across classes* and contamination is not (1.04×–1.37×), and this dataset exists
to compare classes and bands.

**The pile is not rebuilt here.** #3670's build did run (2026-09-06 19:15, 12
classes, 9,900 provable, all five embedders, 15.5 min) and a parallel study's
25-class rebuild overwrote its cells 37 minutes later. Rather than take a third
build, the rebuild is deferred so it happens once with #3588 (**#3698**). Nothing
in the report needed those cells: `draw_negatives` is hash-ranked and
roster-pinned, so `negpool_coverage.py` redraws the exact pool off the loader's
own passes, and its counts reproduce the 19:15 build's `--verify` output exactly.

## Two defects found on the way, both fixed here

**`apply_corrections` folded a verdict for any class**, and `corrections.json`
is shared by every build of this family. #3588's negative pass added thirteen
classes to that file at 19:26 on 2026-09-06, and from that moment a twelve-class
build — the shipped construction on `dev` — died three passes later with
`KeyError: 'car'`. A shared file made the shipped dataset unbuildable and
reported it as a dict lookup.

**`labels_exhaustive` was two claims wearing one name**: set by a COCO anchor,
*and* by any human who looked — at one class. Under `provable` that would have
admitted images to an "all-provable" pool whose absence claim is still VG's
silence for eleven of the twelve. The draw now stratifies on a `coco_scored` set
captured before corrections widen `exhaustive`, stamped per media so `--verify`
checks the strict claim. The strictness is free: 34,071 candidates against 9,900
needed. #3667's `_evaluable` still reads the loose flag — 197 images apart —
which is #3697 rather than a change to what #3667 admits.

Also: `vg_scale_deep` asserted its prevalence against `SCALE_PREVALENCE`, which
this moves. A correct deep rebuild would have spent its GPU hours and then
aborted, naming the wrong constant.

## The gates

Two new invariants in `--verify`, because a construction change with a deferred
rebuild is otherwise silent — the cells load, band and score, and only the
relation to the constants is broken:

- the designated pool must have `SCALE_N_NEG` images, and the failure quotes the
  prevalence the cell actually has;
- under `provable` every designated negative must carry `coco_scored`, and a
  cell predating the stamp is told to rebuild rather than passed on the weaker
  flag.

And the review-coverage gate learns the difference between the two ways an image
leaves the pool. `provable` makes **513 of 743** reviewed negatives ineligible —
the review deliberately over-sampled the off-COCO half (24.9% COCO-scored
against the pool's 45.0%), which is #3675's recommendation working. Charging
those to coverage reports 22% and reads as a catastrophe; forgiving them
silently would let a real reshuffle hide behind the rule. They get their own
column and a line above the verdict, the denominator keeps every image the
composition *can* hold — **coverage of those is 100.0%** — and a gate left
judging fewer than 50 images says so instead of printing a confident 100%.

Closes #3670
Refs #3693, #3681, #3690

https://claude.ai/code/session_01FAnc7gMSFh43SCLGBM3zCv


samggreenberg:
**#3588 (the 25-class expansion) is in flight in a parallel session**, and the two branches touch four of the same pile files. Whoever merges second should expect:

- `check_review_coverage.py` — **the one non-mechanical conflict.** Both PRs add a reason an image may leave the pool without that counting against coverage, by different mechanisms: #3588 records `disqualified` ids in the roster (the class list grew past the image) and folds them into `corrected`; this PR computes eligibility from the composition rule and gives it its own `by rule` column. Both reasons are real and distinct — keep them as separate columns rather than routing the class-list exclusions through `corrected`, which would conflate "the review promoted it" with "the class list grew past it". With both applied at once the `MIN_DENOM` guard here also starts mattering: two independent exclusions can leave a coverage figure quoted off a handful of images.
- `launch_pile.sh` — identical fix on both sides (#3693). Take either.
- `pilebuild/loaders/vg_scale.py` — keep `apply_corrections` skipping verdicts for classes outside *C*, and keep the `coco_scored` capture **before** `apply_corrections` widens `exhaustive`. The first is inert on a 25-class branch, which is what makes it easy to lose in a merge — and it is the fix that makes a twelve-class build possible at all today.
- `pile_config.py`, `tests_lib/meta/test_pile_vg_scale.py` — additive on both sides.

The rebuild stays deferred until both land: #3698.

Issue #3778, opened 2026-09-09
vase: the rule says a planter IS a vase, COCO and the reviewer both say it is not


samggreenberg:
**Difficulty: medium.** A ruling, a rule edit, and a re-review of the planter cases before the next rebuild.

**Recommended Claude model: Sonnet 5.** The measurement is done; the judgement is the owner's.

## The contradiction

`SCALE_CLASS_RULES["vase"]` is named **"vase incl pots and planters"** and says:

> Good: only a vessel MADE as one -- vases, flower pots, planters, urns,
> pottery. [...] **A potted plant's pot is a vase; vote the vessel, not the
> plant.**

The reviewer finished the `vase` slate (116 good, 278 bad) rejecting planters,
and flagged the line as weak -- judging on proportions, ornateness and
indoor/outdoor rather than on anything written. So practice and rule disagree,
and the rule is the string the detector's own name carries (#3612).

## COCO settles it, and it settles it against the rule

`potted plant` is a COCO class in its own right, so this is measurable:

| | |
|---|---|
| COCO images with a `vase` | 3,730 |
| COCO images with a `potted plant` | 4,624 |
| ...of which hold **no** `vase` at all | **69%** |
| potted plants (in images that DO hold a vase) reaching IoU > 0.5 with one | 82 of 2,500 |
| ...overlapping no vase box at all | 1,433 of 2,500 |

COCO does not annotate a planted pot as a vase.

## Why this is not a matter of taste

`vg_scale`'s `vase` **negatives are drawn from the COCO-anchored half** (#3670).
An image COCO annotates as `potted plant` and not `vase` is therefore a
*provable negative* for `vase` in our own pool -- while the shipped rule makes
that same image a positive. The rule contradicts the negatives the dataset is
built from, and the reviewer's practice agrees with the negatives.

This is the `bench`/`chair` shape again (#3756): two parts of the definition
answering one question differently, found mid-pass by the person voting.

## Proposed

Strike "A potted plant's pot is a vase" and narrow the Good list: a vase, urn or
decorative empty pot is Good; **a pot with a plant growing in it is a
`potted plant`, not a vase.** Rename the rule accordingly -- "vase incl pots and
planters" is exactly the claim being withdrawn.

Two things to rule on rather than assume:

1. **an empty flower pot**, no plant. COCO's convention is unclear here and it
   was not measured; the draft keeps it Good.
2. **a window box or a large outdoor planter.** The reviewer's "more outdoors =
[...]

Issue #3768, opened 2026-09-09
Measure what the screening threw away: sample below the cut on the four finished classes


samggreenberg:
**Difficulty: low.** One sampled slate and one reviewer pass; the machinery all exists.

**Recommended Claude model: Sonnet 5.** Sampling and scoring are mechanical.

## Why

The four finished slates measure the screening's **precision** cleanly -- every
candidate was judged, so there is no ranked-prefix bias:

| class | slate | good | bad | precision |
|---|---|---|---|---|
| `dog` | 73 | 59 | 14 | 0.81 |
| `boat` | 141 | 65 | 76 | 0.46 |
| `bicycle` | 37 | 17 | 20 | 0.46 |
| `fire hydrant` | 166 | 47 | 119 | 0.28 |

They say **nothing about recall**, because by construction nobody looks below
the cut. That is the one number the whole design rests on: the cuts were chosen
to hit ~95% recall *on the COCO-anchored half*, and the anchored half is a
different stratum from the queue. Everything downstream -- how much of the
exhaustive debt this pass discharges, and what residual silence rate `vg_scale`
carries -- is unknown until it is measured.

## What to do

For each of the four finished classes, draw a random sample (~100 images) from
the queue images **not** flagged for that class, box nothing, and ask the same
Good/Bad question. The hit rate on that sample estimates the miss rate directly.

- draw from rows the slate excluded **by score**, not the ones excluded as known
  As or as already answered -- those are different exclusions and only the score
  one is the screening's own decision;
- these carry no box, so the question is "is there a `<class>` anywhere in this
  image?" rather than the slate's "is this box a `<class>`?". That deserves its
  own detector name so the two questions are never pooled (#3612);
- 100 per class at a true miss rate of 5% yields ~5 hits. Enough to tell 5% from
  25%, which is the decision at hand; not enough to quote a rate.

## What to be careful about

**A zero-hit sample is not proof of zero misses.** With n=100 and no hits the
upper bound is still ~3%, which over 3,397 images is ~100 lost positives. Report
the interval, not the point.

**These four will not generalise.** Precision already spans 0.28-0.81 across
them and recall probably varies as much. This is a spot check on the method, not
a per-class rate for the other twenty-one.

Related: #3720, #3612. Slates and cuts in
`scripts/experiments/pile/make_class_slates.py`; finished verdicts in
`human_record/LABELSETS__slate__*.json`.

samggreenberg:
[...]

PR #3783, opened 2026-09-10
The recall measurement is complete: 600 below-cut images, zero positives

samggreenberg:
The recall measurement is complete: 600 below-cut images, zero positives

All six below-cut sets are answered (#3768). Each is 100 images drawn at random
from BELOW that class's score cut, shown with no box, asked "is there one
anywhere in this image?":

    bicycle  0/100      fire hydrant  0/100      boat       0/100
    dog      0/100      sink          0/100      stop sign  0/100

600 images, **zero positives found**. This is the number the finished slates
could never produce, and the one the whole screening design rested on: the cuts
were chosen to keep ~95% of positives on the COCO-anchored half, and nothing
until now tested whether that transferred to the queue.

Stated as a bound rather than a zero, which is what 600 samples support:
pooled, the one-sided 95% upper bound on the miss rate is **0.5%**; per class,
with n=100 each, it is 3.0%. Pooling assumes the six behave alike and they
demonstrably do not -- precision across them spans 0.096 to 0.81 -- so the
per-class bound is the honest one to quote and the pooled figure is a summary.

Either way it rules out the failure that mattered. The worry was that screening
15.5% of the sweep might quietly discard a quarter of the positives; it does not
appear to discard any.

Also banks `bench` complete (104g/985b, precision 0.096) and adds
`owl_planter.py`, which sizes the pending `vase` ruling (#3778): 167 of the 278
vase rejects score `potted plant` or `flower pot` above `vase`, against 18 of
the 116 accepts. So the planter question is worth roughly 130-167 images in
`vase` alone -- applying the rule as written would have about doubled its
positives.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_012yKKtCnzocPQy6yZGQyzTZ


samggreenberg:
Superseded. Every script on this branch is already on dev via the later banking PRs, and its verdict files are stale -- `bench` finished at 104g/985b and `fork` has since been added. The one file that never reached dev, `owl_planter.py`, is carried in its own PR. The result this PR reported stands and is recorded in #3768, now closed: 600 below-cut images, zero positives.

PR #3809, opened 2026-09-12
Clear the npm audit gate: Angular 21.2.23/24 lockfile refresh (#3795)

### LINKED ISSUE #3795: npm audit gate is red on dev: 7 moderate Angular advisories against the pinned 21.2.19
**Difficulty: low, but environment-gated.**

**Recommended Claude model: Sonnet 5.** The change itself is a lockfile bump; what it needs is a machine whose npm can actually resolve the newer Angular patch releases, not deep reasoning.

`./run-tests.sh` fails on `dev` for every branch, at the `npm audit` lane only:

```
7 moderate severity vulnerabilities
@angular/common   21.0.0-next.0 - 21.2.19  GHSA-p297-fm68-3q8c  (HttpTransferCache bypass via withRequestsMadeViaParent)
@angular/compiler 21.0.0 - 21.2.19         GHSA-hh8m-fm6v-7cvg  (sanitization bypass via directive host bindings)
```

Everything else is green — `Frontend build OK`, Vitest OK, pytest OK, pyright/pip-audit/vulture OK. This is the same shape as #3529: an advisory published against a version the lockfile pins, turning one gate red for work that has nothing to do with it.

### The fix should be in-range

`frontend/package.json` already declares `^21.2.17` for the Angular packages, and 21.2.23 (`@angular/build` / `@angular/cli`: 21.2.24) is published, so this is a **lockfile refresh**, not a version-range change.

### Why it is not done here

Attempted in the Claude Code on the web container (while working #3794) and could not be resolved there:

- `npm audit fix` — no change, still reports 7.
- `npm update [--prefer-online] @angular/{core,common,compiler,compiler-cli,forms,platform-browser,router}` — `package-lock.json` untouched, tree stays at 21.2.19.
- `rm -rf node_modules && npm install` — resolves 21.2.19 again (the lock pins it).
- `npm install @angular/…@21.2.23 --save` (all nine together, `@angular/build`/`cli` at 21.2.24) — `ERESOLVE`: the peer graph is reported against the *installed* 21.2.19 copies of `forms` / `platform-browser` / `router`, so the whole set cannot move as one.

`npm view @angular/core@21` does list 21.2.21–23 from this container, so the metadata is reachable; the installer just will not move off the lock here. Forcing it (`--force` / `--legacy-peer-deps`) would risk a mixed-version Angular tree, which is not something to slip into an unrelated PR.

### What to do

On a laptop with a normal npm environment: bump the nine `@angular/*` entries to their latest 21.2.x together, commit **`frontend/package-lock.json`** (and `package.json` only if a range actually needs widening), then confirm `./run-tests.sh frontend` is fully green — build, `npm audit`, and Vitest.

If the resolution turns out to be genuinely blocked upstream rather than environment-specific, the fallback is the one #3529 used: record the advisories as accepted with a reason, so the gate reports something true instead of being red for everyone.

samggreenberg:
Fixes #3795.

`./run-tests.sh` failed its `npm audit` lane on `dev` for **every** branch. Two moderate advisories land against the pinned 21.2.19 tree:

- **GHSA-p297-fm68-3q8c** — information leak via `HttpTransferCache` bypass when using `withRequestsMadeViaParent`
- **GHSA-hh8m-fm6v-7cvg** — sanitization bypass via directive host bindings

They fan out to the 7 reported findings across `common`/`forms`/`platform-browser`/`router` and `compiler`/`compiler-cli`/`core`.

`frontend/package.json` already allows the fix (`^21.2.17`, `^21.2.16`), so the bump itself is **a lockfile refresh only — no range changed**.

Three commits, because clearing the audit gate uncovered two things sitting behind it.

---

### Why the obvious commands did not work

The issue could not get past this in the web container and read it as environment-specific. It is not — it is a property of the dependency graph:

**The Angular framework packages peer-depend on each other at an EXACT version**, not a range — `@angular/common@21.2.19` requires `@angular/core@21.2.19`. They form a clique that can only move as one atomic set.

| attempt | result |
|---|---|
| `npm update --package-lock-only @angular/...` (all ten) | moved `@angular/cli` alone; framework stayed at 21.2.19 |
| `npm install --package-lock-only @angular/...@21.2.23` (all nine at once) | `ERESOLVE` — npm anchors the peer graph on the versions **already recorded in the lockfile**, so it still saw `forms@21.2.19` demanding `common@21.2.19` |

The issue diagnosed that second failure as the peer graph being read off the *installed* copies in `node_modules`. It reproduces with **no `node_modules` present at all** — the lockfile alone is enough to pin it, which is why the same command also failed here on an ordinary npm environment.

**What works:** drop the ten `node_modules/@angular/*` entries from the lockfile and let `npm install --package-lock-only` refill them. With no stale entries to anchor on, the whole set resolves in one step.

### Scope of the diff

Deliberately surgical. The only non-`@angular` movement is Angular's own toolchain — `@angular-devkit/*` and `@schematics/angular` 21.2.19 → 21.2.24 — plus `@modelcontextprotocol/sdk`, which `@angular/cli` pulls in.

The alternative (delete the lockfile, regenerate) reaches the *same* Angular versions but rewrites far more. Both were built and compared:

| | Angular result | diff lines | non-Angular bumps |
|---|---|---|---|
| full regeneration | identical | 1229 | 123 |
| **surgical (this PR)** | identical | **134** | **5** |

Final: framework `21.2.23`, `@angular/build` / `@angular/cli` `21.2.24`, `@angular/cdk` unchanged at `21.2.14` (already its 21.2.x tip).

---

## 2. The bundle budget: 520kB → 525kB

With the audit gate green, the **frontend build gate went red in its place** — the patch costs 3.17kB of framework code and `dev` had under 1kB of headroom.

Measured on one cpu node, same build, only `package-lock.json` swapped between runs:

| lockfile | initial bundle | vs 520kB budget |
|---|---|---|
| `dev` (Angular 21.2.19) | **519.04 kB** | 0.96 kB under |
| this branch (21.2.23) | **522.21 kB** | **2.21 kB over** |

`angular.json` says not to raise the budget without first auditing the eager path for genuine bloat, and `CLAUDE.md` / `docs/FRONTEND.md` require **the user's explicit approval**.

- **The audit is the diff**: commit 1 changes `package-lock.json` and nothing else, so no app code entered the eager path — the whole +3.17kB is framework.
- **Approval**: given 2026-09-12 on the two measurements above.
- **525kB, not the minimum 523kB**: `dev`'s headroom was already 0.96kB and this single patch consumed it. 2.79kB restores a comparable margin so the next Angular patch does not re-red the gate on its own. It is also the value the budget held before (540 → 525 → 520).

The standing "audit before raising" instruction is kept in `angular.json`, now scoped to raises driven by app code, with this measurement recorded next to the number.

---

## 3. `check-json` vs the repo's eight JSONC files

Committing commit 2 was **impossible with hooks enabled**:

```
frontend/angular.json: Failed to json decode (Expecting value: line 54 column 17)
```

Not new, and not specific to `angular.json`. Eight tracked files fail a strict parse today, all legitimately JSONC:

```
frontend/.vscode/{extensions,launch,tasks}.json
frontend/angular.json
frontend/tsconfig{,.app,.spec}.json
pyrightconfig.json
```

Angular, TypeScript, pyright and VS Code all document comment support in their config files, and this repo actively **depends** on it — 08c17ca7 moved the bundle-budget rationale *into* a JSONC comment precisely because the Angular schema rejects extra keys. The files are right; the hook is wrong.

It stayed invisible because pre-commit only passes **staged** files: the hook is a no-op until somebody edits one of the eight, and then it blocks them outright. (Which suggests prior edits to these files went in under `--no-verify`.) The exclude keeps `check-json` doing its real job on actual `.json` files.

---

## Verification

GRID, `rack2n10`, against the pushed head — the lockfile A/B was job 642479, the gates job 642481:

- `npm ci` from the new lockfile — clean install, **`found 0 vulnerabilities`**
- `npm ls --depth=0` confirms the *installed* tree is really 21.2.23/21.2.24, not just the lockfile text
- `./run-tests.sh frontend` — **`Frontend build OK`**, **`npm audit: OK`**, Vitest
- `./run-tests.sh` — the full suite, so this is confirmed green everywhere and not only on the frontend lane

https://claude.ai/code/session_01DH8JkPckirmtQM4S91sVnq

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,6 +35,22 @@ repos:
       - id: check-yaml
       - id: check-toml
       - id: check-json
+        # These files are JSONC, not JSON.  Angular, TypeScript, pyright and
+        # VS Code all document comment support in their own config files, and
+        # this repo depends on it: the bundle-budget rationale lives in an
+        # angular.json comment precisely because the Angular schema rejects
+        # extra keys (08c17ca7).  The hook parses with a strict json.load, so
+        # it rejects all eight of them.  That stayed invisible because
+        # pre-commit only passes STAGED files -- editing one of these is rare,
+        # and when it happens the hook blocks the commit outright with
+        # "Failed to json decode" (#3795).  The hook is kept for real .json.
+        exclude: |
+          (?x)^(
+            frontend/\.vscode/.*\.json|
+            frontend/angular\.json|
+            frontend/tsconfig(\.\w+)?\.json|
+            pyrightconfig\.json
+          )$
       - id: check-merge-conflict
       - id: check-added-large-files
         # 4 MB, not the default 500 KB.  This hook exists to keep datasets and
--- a/frontend/angular.json
+++ b/frontend/angular.json
@@ -51,7 +51,7 @@
           "configurations": {
             "production": {
               "budgets": [
-                // initial: 520kB. The app is zoneless (zone.js dropped from the
+                // initial: 525kB. The app is zoneless (zone.js dropped from the
                 // production polyfills), which took the measured eager bundle to
                 // ~488kB. The eager path is framework-dominated (Angular
                 // core+router+forms+http+CDK), not app bloat: routes are lazy,
@@ -67,9 +67,23 @@
                 // limit sits just above the real size with a small headroom.
                 // Do not raise without first auditing the eager path for genuine
                 // bloat.
+                //
+                // Raised 520kB -> 525kB on 2026-09-12 (#3795) to take the
+                // Angular 21.2.19 -> 21.2.23 security patch (GHSA-p297-fm68-3q8c
+                // HttpTransferCache bypass, GHSA-hh8m-fm6v-7cvg sanitization
+                // bypass) that the npm audit gate needs. Audited as instructed
+                // above: that change is a package-lock.json refresh and touches
+                // no app file, so the growth is framework-only, not app bloat.
+                // Measured on one machine, same build, only the lockfile
+                // swapped: 519.04kB at 21.2.19 -> 522.21kB at 21.2.23, +3.17kB.
+                // The old 520kB left 0.96kB of headroom and the patch alone
+                // consumed it; 525kB restores a comparable margin (2.79kB) so
+                // the next patch release does not re-red the gate by itself.
+                // The audit instruction above still governs any raise driven by
+                // APP code, which this one was not.
                 {
                   "type": "initial",
-                  "maximumWarning": "520kB",
+                  "maximumWarning": "525kB",
                   "maximumError": "1MB"
                 },
                 {
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -253,13 +253,13 @@
       }
     },
     "node_modules/@angular-devkit/architect": {
-      "version": "0.2102.19",
-      "resolved": "https://registry.npmjs.org/@angular-devkit/architect/-/architect-0.2102.19.tgz",
-      "integrity": "sha512-cj4tzUMiloLTg5rNf17E8MsvIxCWYoBiBsaj7ns6dgXqT9XCeG+J0TA2t1M+N9uuqfeLd22U/rYoCkADmcircQ==",
+      "version": "0.2102.24",
+      "resolved": "https://registry.npmjs.org/@angular-devkit/architect/-/architect-0.2102.24.tgz",
+      "integrity": "sha512-U/d5l/1rfrNSlbGjMTunyzwxh+2vJce3y0r0tIx/NOFZbnYhq+gx+e33mR6oxI4yk/1Jks67ltvstrJuBbVbYg==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
-        "@angular-devkit/core": "21.2.19",
+        "@angular-devkit/core": "21.2.24",
         "rxjs": "7.8.2"
       },
       "bin": {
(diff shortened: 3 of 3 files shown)


samggreenberg:
## Verification: all gates green

GRID job **642481** (`rack2n10`), against the pushed head `6411ca1b`. `COMPLETED`, exit `0:0`, 8 min wall, MaxRSS 10.5 GB.

```
NPM_CI_EXIT=0
FRONTEND_EXIT=0        # ./run-tests.sh frontend
FULL_EXIT=0            # ./run-tests.sh   (full suite)
SUMMARY npm_ci=0 frontend=0 full=0
```

| gate | `dev` before | this branch |
|---|---|---|
| `npm audit` | 7 moderate | **OK** |
| frontend build | OK — 519.04 kB | **OK** — 522.21 kB (budget 525) |
| Vitest | OK | **OK** |
| pytest | — | **11111 passed**, 6 skipped, 1 xfailed, 1 xpassed |
| pyright / pip-audit / vulture | — | OK |

`npm ci` from the refreshed lockfile reports `found 0 vulnerabilities`, and `npm ls --depth=0` confirms the installed tree — not just the lockfile text — is at 21.2.23 (`@angular/build` / `@angular/cli` 21.2.24).

Follow-up filed as #3811: the eager path sits within ~3 kB of its budget as a matter of course, so each Angular patch tends to re-open this. That wants an eager-path audit on its own account, not another budget raise.

PR #3812, opened 2026-09-12
Eager-path audit: drop FormsModule, 522.21kB -> 483.71kB (#3811)

### LINKED ISSUE #3811: The eager bundle sits ~3kB from its budget, so every Angular patch re-reds the build gate
**Difficulty: medium.** An eager-path audit, not a config change.

Split out of #3795 / PR #3809, where a routine Angular **patch** release (21.2.19 → 21.2.23, taken for two security advisories) grew the initial bundle by **3.17 kB** and turned the frontend build gate red:

| lockfile | initial bundle | vs the 520 kB budget then in force |
|---|---|---|
| Angular 21.2.19 | 519.04 kB | 0.96 kB under |
| Angular 21.2.23 | 522.21 kB | 2.21 kB **over** |

Measured on one cpu node, same build, with only `package-lock.json` swapped between the two runs.

## The problem is the margin, not the number

The budget was raised to 525 kB in #3809 (with approval, and with the audit recorded next to the number) because the growth was framework-only — that change touched `package-lock.json` and nothing else. That unblocks the security patch but does not fix the underlying situation:

- The eager path is **framework-dominated** by design — Angular core+router+forms+http+CDK — which `angular.json` already documents.
- So its size moves with **Angular's** releases, not with our code.
- A 3 kB patch-release drift is therefore normal, and the headroom has been in that same 1–3 kB band for a while (540 → 525 → 520, now 525).

The consequence is that an unrelated, routine dependency bump periodically lands on whoever happens to be holding it, as a red gate that looks like their fault. That is the same failure mode #3529 and #3795 describe, one layer down: a gate that goes red for reasons unconnected to the change under test trains people to shrug at `RUN FAILED`.

## What to do

Audit the eager path for genuine reclaimable weight, rather than raising the budget again:

- The last audit of this kind (splitting `DashboardSortService` out of `DashboardColumnsService`) recovered ~6.5 kB, so there is precedent for real wins.
- `vt-modal` is eager and drags CDK a11y's focus trap (~7 kB) across all ~24 dialogs — worth re-checking whether it still has to be eager.
- Anything reclaimed should go into **headroom**, not into a lower budget number: the goal is that a patch-level framework bump stops being an event.

If the audit finds nothing meaningful left, that is a useful answer too — it would say the budget should track the framework with a deliberate margin (say 10 kB) and a note explaining why, instead of being re-litigated a few kB at a time.

## Worth knowing

`./run-tests.sh` treats **every** `▲ [WARNING]` from `build:prod` as a hard failure, and that strictness is deliberate — fb32d003 fixed the gate's ANSI-blindness precisely because an over-budget bundle had sailed past it for months. Nothing here argues for relaxing that.

samggreenberg:
Fixes #3811.

The issue was scoped around clawing back the ~3 kB that an Angular patch release costs. The audit found **38.50 kB**, in one place.

| | initial bundle | headroom under the 525 kB budget |
|---|---|---|
| `dev` | 522.21 kB | 2.79 kB |
| this PR | **483.71 kB** | **41.29 kB** |

Measured on one cpu node, production build, nothing else changed between runs.

## What the audit found

Attributing every eager byte back to its source module (esbuild metafile, walking **only** static `import-statement` edges from `src/main.ts`, so `@defer` and lazy routes are correctly excluded) gives:

| | before | after |
|---|---|---|
| framework (npm) | 354.50 kB | 317.54 kB |
| app code | 121.58 kB | 121.47 kB |

So the eager path is ~74% framework, and the app's own contribution was **already lean** — the last audit had done its job. The weight was one npm package that had no business being eager:

```
src/main.ts
  -> src/app/app.component.ts
  -> src/app/components/dialog-host/dialog-host.component.ts
  -> @angular/forms          36.96 kB
```

`vt-dialog-host` is rendered unconditionally by `AppComponent`, and `vt-login` sits on the same eager path. Between them they imported `FormsModule` for **two uncontrolled text fields** — the login username box and the `prompt()` input.

Both now use plain `[value]` + `(input)`; the login form swaps `(ngSubmit)` for native `(submit)` with an explicit `preventDefault()`. **Nothing is deferred**, so no interaction pays a lazy-chunk fetch — the bytes leave the eager graph outright rather than moving to another chunk. Verified after the change: `@angular/forms — eager? no`.

The login form had no other dependency on `NgForm`: no `#f="ngForm"` reference, and no `required`/`pattern`/`type=email`, so the `novalidate` that `NgForm` silently adds had nothing to suppress. Both components' existing specs already drove the inputs through native `input`/`submit` events, so they cover the new implementation unchanged.

## Left eager on purpose

`@angular/cdk` a11y (~9 kB), reached via `vt-modal` inside the always-rendered `vt-dialog-host`. Dropping it costs the focus trap on all ~24 dialogs; deferring the dialog host puts a chunk fetch in front of every `confirm()`/`prompt()`. Neither is worth 9 kB when there is now 41 kB of headroom, so the trade is written into `angular.json` instead of being rediscovered next time.

## The budget can no longer police this

That is the uncomfortable half of the result: with 41 kB of slack, a re-added `FormsModule` would fit **silently**, spending exactly the margin that exists so a framework bump does not turn the gate red for everyone (#3795). Per the issue, the reclaimed bytes stay as headroom rather than becoming a lower number — so the specific trap gets a test instead.

`TestEagerBundleComposition` resolves the eager chunk set from the build output: `main.js` plus everything reachable through `from"./chunk-X.js"`, **never** following `import("./chunk-Y.js")` — that second form *is* the `@defer`/lazy-route boundary, and following it would walk the whole app and make the assertion vacuous. It then asserts `ngNoForm` is absent; that string is from `NgForm`'s own selector (`form:not([ngNoForm])`) and appears in exactly one chunk of the build, so it is a precise marker.

**The guard was verified to fail, not assumed to work.** With the first commit reverted and the bundle rebuilt at 522.21 kB it fails; with the fix it passes:

```
PHASE A (FormsModule eager, 522.21 kB):  1 failed, 1 passed   <- correct
PHASE B (fix in place,      483.71 kB):  2 passed             <- correct
```

A second test asserts the walk still resolves a plausibly-sized bundle, so if the build output's import syntax ever changes, that surfaces as a failure rather than a guard that silently matches nothing.

## Verification

GRID, full `./run-tests.sh`: **11115 passed**, 6 skipped, build OK, `npm audit` OK, Vitest OK.

(The 2 `xpassed` are the pre-existing pair in `tests_lib/cli/test_tqdm_progress.py`, marked *"partialmethod identity is unstable across attribute accesses"* — self-describing run-to-run flakiness in tqdm code, unrelated to anything here.)

https://claude.ai/code/session_01DH8JkPckirmtQM4S91sVnq

### DIFF
--- a/docs/FRONTEND.md
+++ b/docs/FRONTEND.md
@@ -726,6 +726,21 @@ written into `angular.json` next to the number. `./run-tests.sh` treats **any**
 `anyComponentStyle` budget. Fix the bloat — split the component, extract shared
 styles, delete dead rules. Raising a budget needs the user's explicit approval.
 
+What actually sits in the eager bundle is mostly **framework**: 317 kB of
+Angular against 121 kB of app code when it was last audited (#3811). That is
+why the budget carries visible headroom — Angular's own patch releases move the
+number a few kB at a time, and when the margin was under 1 kB a routine
+security bump turned the build gate red for every branch (#3795). Treat the
+headroom as reserved for the framework, not as room for app code.
+
+The corollary is that the budget cannot catch an app-code regression on its own
+any more, so one specific trap is covered by a test:
+**do not import `FormsModule` into anything on the eager path.** `vt-login` and
+`vt-dialog-host` each render a single uncontrolled text field and used to pull
+all ~37 kB of `@angular/forms` for it; both now use plain `[value]` + `(input)`.
+`TestEagerBundleComposition` in `tests/core/test_frontend.py` walks the built
+bundle's *static* imports from `main.js` and fails if the package reappears.
+
 ### Other primitives
 
 `components/icon/` maps names (and backend-supplied emoji) to sanitised inline
--- a/frontend/angular.json
+++ b/frontend/angular.json
@@ -51,36 +51,36 @@
           "configurations": {
             "production": {
               "budgets": [
-                // initial: 525kB. The app is zoneless (zone.js dropped from the
-                // production polyfills), which took the measured eager bundle to
-                // ~488kB. The eager path is framework-dominated (Angular
-                // core+router+forms+http+CDK), not app bloat: routes are lazy,
-                // the heavy modals are @defer'd, and `marked` lives only in a
-                // deferred chunk. The shared `vt-modal` adds CDK a11y's focus
-                // trap (`CdkTrapFocus`) for keyboard/screen-reader focus
-                // management across all ~24 dialogs; because the modal is eager,
-                // that ~7kB lands in the initial bundle, taking it to ~507kB.
-                // Measured 515.63kB as of the eager-path audit that split
-                // `DashboardSortService` out of `DashboardColumnsService` (the
-                // eager context pulldowns were dragging `ManagedColumns` and the
-                // Dashboard column metadata onto first paint, ~6.5kB). The warn
-                // limit sits just above the real size with a small headroom.
-                // Do not raise without first auditing the eager path for genuine
-                // bloat.
+                // initial: 525kB. The app is zoneless (zone.js dropped from
+                // the production polyfills). The eager path is
+                // framework-dominated, not app bloat: routes are lazy, the
+                // heavy modals are @defer'd, and `marked` lives only in a
+                // deferred chunk. Measured 2026-09-12 (#3811) the eager split
+                // is 317.54kB framework to 121.47kB app code.
                 //
-                // Raised 520kB -> 525kB on 2026-09-12 (#3795) to take the
-                // Angular 21.2.19 -> 21.2.23 security patch (GHSA-p297-fm68-3q8c
-                // HttpTransferCache bypass, GHSA-hh8m-fm6v-7cvg sanitization
-                // bypass) that the npm audit gate needs. Audited as instructed
-                // above: that change is a package-lock.json refresh and touches
-                // no app file, so the growth is framework-only, not app bloat.
-                // Measured on one machine, same build, only the lockfile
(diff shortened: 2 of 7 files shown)


samggreenberg:
## Verification: green on the pushed head

GRID job **642496** (final suite, run after the `docs/FRONTEND.md` addition so the doc gates are covered too). `COMPLETED`, exit `0:0`, 6m38s.

```
All checks passed!          # ruff, codespell, check-docs, openapi snapshot, …
Frontend build OK           # 483.71 kB, no budget warning
npm audit: OK
ALL 11115 TESTS PASSED (6 skipped, total: 11121)
FINAL_EXIT=0
```

Guard validity was proven separately in job 642491 rather than assumed:

```
PHASE A  components reverted, rebuilt at 522.21 kB  ->  1 failed, 1 passed   (guard fires)
PHASE B  fix in place,        rebuilt at 483.71 kB  ->  2 passed             (guard clears)
```

Two notes on the counts, so neither reads as a surprise later:

- **11117 → 11121 collected.** Two are the new guard tests; the other two came from #3810 and #3805, which peers merged to `dev` between this branch's base and the earlier #3809 run.
- **2 `xpassed`.** Pre-existing, in `tests_lib/cli/test_tqdm_progress.py`, both marked *"partialmethod identity is unstable across attribute accesses"* — they flip run to run by their own description (the #3809 run saw 1 xfail + 1 xpass). Nothing to do with this change, which touches no Python outside `tests/core/test_frontend.py`.

Issue #3720, opened 2026-09-08
Emit the annotation queue for vg_scale's off-COCO positives — the pass cannot start without a worklist


samggreenberg:
`docs/plans/vg-scale-exhaustive-annotation.md` scopes the move from a designated
benchmark to an exhaustively annotated set, and its first open item is the one
nobody can start without: **the worklist**. A reviewer cannot begin the pass
until they know *which* images are off-COCO positives, and that is a join
between VG and COCO ids — nothing about the image says it.

**Recommended Claude model: Sonnet 5.** The rule is decided and written down;
the work is one script, its tests, and one run against the shipped cell.

## What is owed

A script under `scripts/experiments/pile/` that emits the queue of images the
exhaustive pass has to cover, plus the per-class counts that fall out of
building it.

The selection rule is already in the data:

* Where the shipped cell carries the `coco_scored` stamp, the queue is
  `categories and not coco_scored` read straight off the pickle — `categories`
  is what an image was designated a positive for, and `coco_scored` says COCO
  answered for all eighty classes at once. The current cells were rebuilt on
  2026-09-07 from `8a62332` and should carry it.
* Where a cell predates the stamp, the fallback is `coco_anchor/image_data.json`'s
  `coco_id` field (~100 MB, already staged) joined against the pickle's
  positives. Note the direction of its error: that pairing is a slight
  *superset* of what the build anchors (it does not re-apply the aspect-drift
  filter that drops 49 of 51,497 pairs), so the fallback can call an image
  anchored that the build did not — which *drops* an image that needs
  annotating. `check_review_coverage.py` uses the same join in the opposite
  direction, where a superset is the safe side. Say which path produced each
  row, and report how much of the queue rests on the fallback.

Neither path needs `objects.json` or a compute node.
`scripts/experiments/pile/exact_supply.py` does, which is why it is the wrong
tool here.

## What it is not

Not a measurement of the queue. The plan says the ~3,100–3,200 estimate does not
need refining — the work has to happen whatever the number is — and the exact
per-class counts arrive **for free** as a byproduct of the worklist. Do not run
anything extra to produce them.

## Done when

The queue has been emitted against the shipped `vg_scale` cell on the pile, with
one row per image (deduped — a positive can be designated in several cells, and
the pass is one exhaustive judgement per image, not one per cell), carrying what
a reviewer needs to open it and what a scorer needs to join it back:
the VG image id, its path, the cells it is a positive for, and which of the two
paths above answered.

[...]

Issue #3612, opened 2026-09-03
cell phone rule: the "base station" clause also rejects a docked mobile


samggreenberg:
Found while reviewing the first #3588 slate (`cell phone not landlines`, 300/300 voted).

The rule's test is *"a handheld slab with a screen and no cord is Good; anything with a cord
or a base station is Bad."* The base-station clause exists to exclude landlines, which is the
class's whole risk (VG `phone` lands on no COCO class 46.2% of the time, worse than `book`'s
43.3%). It works: three of the five rejected pre-boxed positives are unambiguous landlines
(2405860 hotel desk phone, 2328733 office desk phone, 2316765 cordless handset on its base).

But it over-triggers. **2387021** is a mobile phone sitting in a charging dock. It is a cell
phone by every reading of the class, and the test as written says Bad because there is a base
under it. The clause is discriminating on **a base being present*, when what it means is
*the handset is not itself the whole device*.

Suggested wording, keeping the landline intent:

> Bad if the handset needs the base to work — landline handsets, desk phones, payphones,
> wall phones, intercoms. A mobile phone resting in a charging dock or cradle is still Good.

This matters beyond one image because the rule string travels in the dataset name and is the
only thing a reviewer sees while voting (which is exactly why `book` split). Any re-review of
this class should use the corrected wording.

Related: #3588, #3605.

samggreenberg:
Addressed in #3622.

The corrected wording is recorded verbatim as `pile_config.SCALE_CLASS_RULES["cell phone"].test`, and `review_name()` is what all four slate makers now build their reviewer-facing dataset/detector name from — the first pass included, not just the re-review, since a rule that only reaches a re-review arrives after the verdicts it exists to prevent.

The table records the short `name` (`cell phone not landlines`) and the full `test` separately, because this class split on the second while the first was fine.

---
_Generated by [Claude Code](https://claude.ai/code)_

samggreenberg:
Shipped to main in the 2026-09-09 release — fixed in #3622.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3813, opened 2026-09-12
Retire the 21 planter positives the recheck rejected (#3778)

samggreenberg:
Retire the 21 planter positives the recheck rejected (#3778)

The planter ruling shipped in #3784 and changed what a `vase` and a `bowl`
ARE, which left 80 rows in `corrections.json` cast under a definition that is
now withdrawn. `make_class_recheck.py` put all 80 back in front of the reviewer
as an image-level question -- *does this photo contain one under the new rule?*
-- and 21 came back Bad. This is the other end of that pass: those 21 rows
become `present: false`, so the images leave the class they can no longer be
positives for.

`apply_recheck.py` does it, and what it refuses is the point:

* **a verdict with no row is a refusal, not an insert.** The slate was built
  from the present rows of `corrections.json` itself, so every verdict in it has
  a row to land on; one that does not means the slate and the file have drifted,
  and inventing membership from a question nobody asked about that image is the
  wrong reading of it;
* **the rule in force is checked against the one the reviewer voted under.** The
  labelset records it, because that name is the only wording a reviewer ever
  sees (#3612). If `SCALE_CLASS_RULES` has moved on since, these verdicts answer
  a superseded question -- which is this issue's own failure mode pointed at
  itself, and the one thing here that *is* mechanically checkable;
* a re-confirmation of a row that already says absent, one image answered both
  ways, and a labelset of another `kind` are each a refusal too. A partial apply
  improves none of them.

A retired row carries its own `source` (`human_recheck`) rather than
`human_reject+adjudicated`: nothing was wrong with these rows when they were
cast. The reviewer was right and the definition moved, and an audit that cannot
tell those apart reads a ruling as a pile of annotator errors.

### The `present: false` path had never really run

865 of the live file's 872 rows say `present: true`, so the half of
`apply_corrections` that RETIRES a positive was exercised only by unit tests --
and this lands 21 of them at once, 20 of them `vase` positives sitting in
designated cells with the roster pinning them there.
`TestRetirementLeavesTheCell` walks the three passes that have to agree:

* a retired positive loses its **pinned** seat (a pin is kept only while the
  image is still eligible, and it is not in the supply at all) and the seat is
  backfilled by rank rather than left empty;
* it becomes a **usable negative** -- what a rejection usually cannot buy, since
  a human looked at the photo and said there is none. That is the difference
  between `present: false` and a boxless `present: true`, which lands in
  `unbanded` and is neither;
* it retires **the class, not the image**. `bowl` and `vase` share images in
  this very pass -- `make_class_recheck` symlinks one file into both datasets --
  so popping the image would delete a positive nobody rejected.

### Measured, not assumed

`replay_selection.py` against the live pile, after the write: **21 positives
dropped and 21 added over 4 cells** -- `vase@medium` 15, `vase@small` 4,
`vase@large` 1, `bowl@large` 1 -- and **0 negatives moved**. That is the
retirement list cell for cell with nothing else touched. The retired images are
now eligible for the negative pool and do not enter it at this rebuild, because
the pool is roster-pinned and already full.

It moves membership, so it lands with the next rebuild. `corrections.json` on
the pile is written under the existing lock and a pre-change copy is at
`/exp/sgreenberg/backups/corrections-pre-3778-20260912-1708.json`; the committed
copy is re-exported by `verdict_store.py`.

Refs #3778, #3784, #3720.

https://claude.ai/code/session_019b1xA9ArRkoQscwxTenJE5

### DIFF
--- a/scripts/experiments/pile/apply_recheck.py
+++ b/scripts/experiments/pile/apply_recheck.py
@@ -1,0 +1,220 @@
+#!/usr/bin/env python3
+"""Apply a recheck pass's verdicts to `corrections.json` (#3778).
+
+A ruling changes what a class *is*, and every row cast under the old wording
+then means something the table no longer says. The planter ruling (#3784) is the
+first one to have done that to rows already on disk: `vase` claimed "flower pots,
+planters" and said in as many words that a potted plant's pot is a vase, so 31
+`vase` rows and 49 `bowl` rows were cast under a definition that is now
+withdrawn. `make_class_recheck.py` put all 80 back in front of the reviewer as an
+image-level question -- *does this photo contain one under the new rule?* -- and
+this is the other end of that: 21 of them came back Bad, and a Bad here is a
+positive that has to leave the class.
+
+**A recheck re-answers an existing row and never writes a new one.** The slate
+was built from the present rows of `corrections.json` itself, so every verdict in
+it has a row to land on. An image with no row is therefore not a new correction
+to add, it is evidence that the slate and the file have drifted apart -- so it
+refuses rather than inventing membership from a question nobody asked about this
+image.
+
+**The rule in force is checked, not assumed.** The labelset records the rule
+whose name the reviewer was voting under (#3612: that name is the only wording a
+reviewer ever sees). If `SCALE_CLASS_RULES` has moved on since, these verdicts
+answer a superseded question and applying them would silently apply the old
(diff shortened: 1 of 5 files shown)


samggreenberg:
Full suite green on the GRID (job 642519, 8 workers): **11,128 passed**, 6 skipped, 2 xpassed in 5:31, `PYTEST_EXIT=0`. The eight pre-pytest gates and `deptry` pass too.

Issue #3840, opened 2026-09-14
The trajectory A/B resolves about +-0.004 of cost at 114 cells, and we find that out per study


samggreenberg:
**Difficulty: medium.**

## What happened

#3825 ran four 84-cell trajectory grids. Two of the arms differ by a factor of **fourteen** in how much haystack they move — `ll1e-6` shifts the admitted set with a p90 of 0.70% of the haystack, `ll1e-8` with a p90 of 0.050% — and the A/B could not tell them apart:

| arm | p90 admitted-set move | Δcost (114 paired cells) |
|---|---|---|
| `ll1e-6` | 0.70% | +0.0044 ± 0.0031 |
| `ll1e-8` | 0.050% | +0.0049 ± 0.0037 |

So the instrument bottoms out somewhere around **±0.004 of cost** at this many cells, and below that the *gate* is the tighter measurement rather than the A/B. #3585 hit the same floor from the other side (Δcost −0.0054 ± 0.0050) and read it as "resolves nothing, in the candidate's favour".

## Why it matters

That floor is coarser than several shipped decisions have turned on. The anchor-mass sweep shipped on −0.026 vs pure x-cal; #3547 and #3319 both argued from effects in the 0.005-0.01 range. Each study rediscovers the floor at the end, when the grids have already run and the arms have already been chosen — which is the expensive order.

## What would close it

The cells-to-resolution curve, from **cells that already exist**: subsample the completed paired-cell frames of #3585 and #3825 at k = 8, 16, 32, 64, 114 and plot the standard error of Δcost against k. That gives a number to *design* a grid against ("to resolve 0.002 you need N cells") instead of one to apologise with, and it costs re-analysis rather than a run.

Two things worth reading off the same frames while there:

- **Per-environment variance.** `caltech101_m` cells are near-saturated and contribute almost no Δ; if most of the noise is one or two environments, a grid can be weighted rather than enlarged.
- **Cells vs steps.** The pairing unit is (category, seed) because the two arms vote on different items. Whether more seeds or more categories buys more resolution per cell-hour is not obvious and is answerable here.

## Why this needs a run

The subsampling is re-analysis, but validating the curve's prediction needs at least one grid run at a size it predicts — otherwise it is a bootstrap estimate checked against itself.

Refs #3825, #3585, #3547, #3319.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BNtgcGLYqW3RJKxcPRQmJt

[...]

Issue #3839, opened 2026-09-14
The anchored refit still exits on max_iter on 7.4% of folds, and the parameter criterion is unreachable


samggreenberg:
**Difficulty: medium.**

#3825 measured the shipped anchored refit and found it leaves on `max_iter` rather than converging on **26.5%** of 4,493 real fold refits (median 113 iterations, p90 at the 200 cap). Moving the stopping rule to the log-likelihood at 1e-8 cuts that to **7.4%**, which is much better and is not zero.

The sharper finding is the control. `iter400` — the incumbent's own parameter rule at 1e-8 given **twice** the budget — still exits on its cap **10.7%** of the time. So on those folds a parameter delta of 1e-8 is not a tolerance the estimator reaches slowly; it is one it does not reach. Doubling the budget again would not fix it.

## What is unresolved

A fit that exits on its iteration cap is not the estimator anyone specified, and after #3825 we can *see* it happening (`FoldAnchoredCut.n_unconverged`, the `fold_anchored_maxiter{u}[a/k]` provenance, and a warning from `_fused_threshold`) without having decided what it *means*. Three options, none costed:

1. **Accept and document a rate.** Pick a tolerance whose non-convergence rate is acceptable and say why, so the number has a target instead of only a monitor.
2. **Give the loop a guarantee.** A relative rather than absolute tolerance, or a stall detector (stop when the objective's improvement per iteration falls below a fraction of its total climb), would make "converged" reachable on a barely identified mixture.
3. **Treat non-convergence as a degeneracy.** `fit_anchored_score_gmm` already falls back to the unanchored fit for inverted means and collapsed components. It is not obvious that a fit which ran out of iterations is *worse* than the unanchored fit, and #3825 has the corpus to settle it — but did not, because it is a different question from where to stop.

## Why this needs a run

Each option is an arm over the captured corpus plus, for anything that moves the cut, the admitted-set gate and a trajectory A/B — the same three instruments #3825 used.

## Where the harness is

`scripts/experiments/gmm_init/` — `arms_3825.py` takes an arbitrary `Stop(max_iter, tol, loglik_tol, anchored_objective)`, and the 84-cell corpus at `/expscratch/sgreenberg/gmm-3585/corpus` needs no re-running. A relative-tolerance or stall-detector arm is one function.

Refs #3825, #3585, #3558.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

Issue #3585, opened 2026-09-02
fit_score_gmm's sklearn GaussianMixture is now the dominant threshold cost — and it is ~12x slower per EM iteration than our own loop


samggreenberg:
**Difficulty:** medium — the engineering is easy and the *gate* is the whole task. Cannot be closed without measuring whether the threshold moves.

**Recommended Claude model: Opus 4.8.** The code change is small; what is hard is establishing that a reinitialised fit is safe, on a function that sets the threshold for *both* the calibration path and every cosine/text sort. A wrong-but-plausible "it's within tolerance" here silently re-prices every detector.

> [!IMPORTANT]
> **Corrected by measurement — read [this comment](https://github.com/samggreenberg/VTSearch/issues/3585#issuecomment-5652965888) before starting.** Three changes to what is written below:
> - The `covariance_type="spherical"` option floated at the end of "The gate" **passes the gate and is 3% slower** — bit-identical cuts on 300 fixtures, zero admitted-set changes, 0.97× speed. It is not the answer.
> - The diagnosis below blames the `"full"`-covariance machinery. Measured, the cost is **EM iterations** (68% at 50k scores, 85% at 7.7k), not the Cholesky path, and `n_iter_` is 26 rather than 12–13.
> - The payoff is **larger** than this body claims: `fit_score_gmm` is **91–95% of a whole cosine/text sort**.
>
> Live candidates, all still behind the gate: `init_params="k-means++"` (5.0–6.3×), a hand-written 1-D EM, and `_GMM_MAX_SAMPLES` (per-iteration cost scales with n).

## What

#3558 made the anchored EM loop 4–6× cheaper, bit-for-bit. That leaves `fit_score_gmm` — sklearn's `GaussianMixture(n_components=2, random_state=42)` — as the dominant remaining term, and it is now the *majority* of an anchored fit:

| free scores | `fit_anchored_score_gmm` | of which sklearn init | of which our EM loop |
|---|---|---|---|
| 7,747 | 0.052 s | **0.039 s (76%)** | 0.012 s |
| 20,000 | 0.220 s | **0.192 s (87%)** | 0.028 s |
| 50,000 | 0.368 s | **0.308 s (84%)** | 0.060 s |

(min of 7, one cloud container; the 20k row's kmeans is anomalously slow there, so read the shape, not the absolute.)

[...]

PR #3841, opened 2026-09-14
Stop the anchored refit on its own likelihood, and report when it does not converge (#3825)

### LINKED ISSUE #3825: The anchored EM is now the whole cost of a fold's fit, and it exits on max_iter rather than converging
**Difficulty: medium** — the code change is three lines; the gate is the task, and unlike #3585 this one moves the shipped threshold with nothing in front of it.

**Recommended Claude model: Opus 4.8.** Same shape as #3585: small diff, and the whole job is establishing that a re-stopped EM is safe on the estimator that sets every trained detector's line.

## What

#3585 replaced the *unanchored* fit and, in doing so, measured the loop next to it. On real fold haystacks captured from the shipped path (`scripts/experiments/gmm_init/`, corpus on `/expscratch/sgreenberg/gmm-3585/corpus`), at n ≈ 2,000 free scores and the shipped κ = 0.3:

| | seconds per fold | EM iterations |
|---|---|---|
| unanchored init (`fit_score_gmm`, post-#3585) | 0.002-0.004 | ~10-25 |
| `_anchored_em` | 0.021-0.048 | **97-200** |

`200` is `max_iter`. On 2 of the 5 folds sampled the anchored loop **did not converge at all** — it ran out of iterations — and the others stopped just inside the cap. So after #3585 the anchored refit is ~90% of a fold's fit, and most of that is EM crawling a flat ridge toward a tolerance (`tol=1e-8`, on the parameters) the sample cannot identify.

## Why the fix is already written down

#3585's finding transfers exactly. The parameter-delta criterion cannot tell *still converging* from *converged, now drifting*, and on a barely-identified mixture it spends all of its time in the second case. Switching the unanchored fit to the log-likelihood rule — sklearn's own, stop when an iteration improves the mean log-likelihood by less than 1e-3 — made it 6-8x cheaper while moving the admitted set by a fraction of what re-initialising the incumbent moves it.

`_anchored_em` **already takes** an opt-in `loglik_tol` for precisely this; the anchored path passes nothing today, which is what keeps it bit-for-bit what #3558 pinned. So the change is: pass it, pick the tolerance by measurement, and gate.

## Why this is not "do the same thing again"

The anchored fit **is** the shipped threshold — there is no second estimator between it and the cut, and no `snap` upstream of it to absorb a tolerance-level move. #3585's own gate found the fold path far more forgiving than the sort path, but that was with the anchored refit *unchanged* and re-converging from whatever init it was handed. Move the refit itself and that cushion is gone.

So this needs the same three things #3585 needed, in the same order: the admitted-set gate over the captured corpus, the log-likelihood comparison (which fit is better, not just different), and the trajectory A/B — because the threshold feeds Autopilot's Hard pick and two arms diverge in *which items get voted on*.

## The harness exists

- corpus: 84 cells of real `fit_fold_anchored_cut` inputs, already captured; no cell needs re-running.
- `gate_3585.py` replays arms that swap `fit_score_gmm`. This needs an arm that swaps **the loop's stopping rule** instead — one function in `arms_3585.py` plus a `swap_fit`-shaped context manager for `_anchored_em`.
- `run_cells_arm_3585.py` + `analyze_ab.py` give the trajectory A/B for free once the arm exists.

## Also worth deciding here

A fit that exits on `max_iter` is not the estimator anyone specified, and nothing reports it. Whatever the stopping rule ends up being, the fold-anchored provenance should be able to say *this fold did not converge* — today it cannot, and the only reason we know it happens is that this measurement went looking.

Refs #3585, #3558, #3166.

samggreenberg:
Closes #3825.

#3585 made the unanchored initialiser 5x cheaper and, timing the loop next door,
found that `_anchored_em` — the refit that **is** the shipped threshold — had
become ~90% of a calibration fold's fit, running a parameter-delta criterion at
1e-8 that on real folds it often never met. The issue proposed #3585's own fix
(stop on the log-likelihood at sklearn's 1e-3) and warned that the gate would be
harder here, because there is no second estimator and no `snap` between this
loop and the green/red line.

The warning was right and **the proposed tolerance is a regression.** The
criterion ports; the number does not.

## What the measurement says

**The incumbent is in worse shape than the issue's five-fold sample showed.**
Over 4,493 real fold refits: median **113** iterations, p90 at the **200** cap,
**26.5%** exiting on `max_iter`, refit is **92.9%** of a fold's fit. And
`iter400` shows a 1e-8 parameter delta is not *slow* on those folds but
**unreachable** — at twice the budget it still exits on its cap 10.7% of the
time.

**1e-3 costs +0.026 ± 0.006 of cost** over 114 paired trajectory cells (4.2
standard errors; FPR +0.044 for FNR −0.018). The worked case shows why: at that
tolerance the refit stops after **10** iterations with both component means
still bunched at 0.34–0.36, *before* the minority component has migrated out to
the high mode at 0.436 where 140 iterations put it. The migration is worth
0.025 nats over 130 iterations, so a 1e-3 rule cannot see it happening. Admitted
set: 70 → 925 of 2,031. This is the opposite of the flat ridge #3585 found — the
loop looks flat here *while it is still doing the work*.

**A second axis the issue did not name decided the design.** `_anchored_em`'s
opt-in `loglik_tol` watched the **free sample's** likelihood, which is right for
`fit_score_gmm` and is not the objective an anchored EM ascends. Traced
iteration by iteration on real folds it **falls on 36.6% of iterations**; the
weighted semi-supervised objective never falls (0 of 1,989). The shipped rule
now watches the latter. The two are identical arithmetic without anchors, so
`fit_score_gmm` is untouched **bit for bit** — asserted in a test, not argued.

**Ships `_ANCHORED_EM_LOGLIK_TOL = 1e-8`**, numerically the incumbent's own
tolerance applied to the likelihood instead of the parameters:

| | value |
|---|---|
| refit cost | **1.9x** cheaper at fold sizes (1.6x at 20k, 1.4x at 50k) |
| admitted set moves | 34.8% of 2,258 cases, **median 0**, p90 **one media in two thousand** |
| worst move in the corpus | **0.24%** of a haystack — against **35.6%** for merely letting the incumbent *finish* |
| anchored fits lost | **0** |
| non-convergence | **26.5% → 7.4%** |
| trajectory A/B | Δcost +0.0049 ± 0.0037 (at the instrument's floor) |

**Plus the half with no trade-off in it: a fit that ran out of iterations can now
say so.** `stats` carries `n_iter`/`converged`/`loglik`;
`FoldAnchoredCut.n_unconverged` counts the folds; the provenance names them as
`fold_anchored_maxiter{u}[a/k]` — the `[a/k]` group kept last and unchanged
because `folds_used` parses it with an end-anchored regex; `_fused_threshold`
logs it. The shipped estimator has been failing to converge on a quarter of
folds, and the only reason anyone found out is that a cost measurement went
looking.

## Two things I want a reviewer to push on

- **No early-stopping arm beat the incumbent's objective once in 4,493 folds**
  (`obj_better` = 0 for all eleven), while `iter400` beats it on 597 and loses on
  none. The incumbent is under-converged *and still climbing*, so this is a
  bounded concession, not a free win. The case for 1e-8 is that 1.7e-8 nats and a
  median midpoint move of 3.7e-5 are below what the quantile-and-snap chain
  resolves — not that it fits better.
- **The A/B could not separate `ll1e-6` from `ll1e-8`** despite a 14x difference
  in how much haystack they move, so its floor is ~±0.004 of cost at 114 cells.
  For these arms the gate is the tighter instrument. `ll1e-8`'s ΔFNR
  (+0.0081 ± 0.0028) is the one sub-metric outside twice its error and is called
  out in §5 rather than left to the cost column.

## Two defects in my own harness, both caught, both in the report

- **The estimator frame was fitting at κ=10**, `fit_anchored_score_gmm`'s bare
  default, while the shipped path uses `FOLD_ANCHOR_WEIGHT` = **0.3**. The
  decision frame always went through the shipped function and was correct;
  nothing crashed, and the numbers were wrong in the flattering direction (43
  iterations and 5.7% non-convergence, against the real 113 and 26.5%). Caught by
  building the worked case and finding its arithmetic did not reconcile.
- **`_anchored_em` was refactored mid-run** (helpers lifted out for the
  complexity gate) while the gate and A/B grids were in flight. Rather than argue
  the refactor was arithmetically identical, the gate was re-run and compared row
  by row: **27,096 rows, 12 shared arms, zero differences**.

## Testing

- `tests_lib/sorting/`, `tests_lib/detectors/`: **2,509 passed**
- All eight pre-pytest gates + `deptry` clean; `selftest_analyze_3825.py` green
- Full suite: job 649454 on a GRID cpu node
- Corpus **re-used, not re-run**: 84 cells of #3585's captures answered a
  question that did not exist when they were taken

Report: `docs/experiments/2026-09-13-anchored-em-stop-3825/REPORT.md`
Follow-ups filed: #3839 (non-convergence still 7.4%, and the criterion is
unreachable), #3840 (the A/B's ±0.004 resolution floor)

https://claude.ai/code/session_01BNtgcGLYqW3RJKxcPRQmJt


samggreenberg:
**Full suite green** on `2bd0cba7b`: 11,392 passed, 0 failed, 6 skipped; pyright, pip-audit, vulture, npm audit, Vitest and all eight pre-pytest gates OK (job 649477, 4m56s on a GRID cpu node).

Two things the first suite run turned up, both resolved:

- **21 pyright `reportOptionalMemberAccess` errors** in the new tests — `fit_fold_anchored_cut`, `fit_score_gmm` and `_anchored_em` all return `T | None`. Fixed in `2bd0cba7b`; the five init sites go through one `_init_fit` helper rather than scattered asserts.
- **One pytest failure, `test_parked_reader_is_woken_by_an_update_from_another_thread`** — an SSE thread-wakeup test untouched by this branch. Treated as a flake only after checking: 3/3 pass on clean `dev` at `dd8dccf68`, 5/5 on this branch standalone, and it passed on the re-run of the same commit. It failed only inside the full xdist run, which is the load condition these timing tests flake under.

`7ac3f561c` adds the incident-log entry for the harness defect described in the PR body (the estimator frame fitting at κ=10 instead of the shipped κ=0.3), since that is the kind of mistake `scripts/experiments/LESSONS.md` exists to stop recurring.

https://claude.ai/code/session_01BNtgcGLYqW3RJKxcPRQmJt

Issue #3853, opened 2026-09-14
Rare 6-20s stall after voting: centre panel black while the next image loads


samggreenberg:
## Symptom

While labeling a VG_Scale slate at a ~2s/click cadence, occasionally the vote registers, **the centre panel goes black**, and the next image takes **6-20s** to appear. Rare but recurring; observed repeatedly on 2026-09-14 (6s, and earlier 10-20s). Not every inter-vote pause is a glitch (reviewer think-time), but every glitch presents as one.

---

# Current state — read this before the comments (updated 2026-09-16)

**Recommended Claude model: Opus 4.8.** Not for the code — for the reading. This thread has already had to retract three confident conclusions (a lifetime NFS counter mistaken for steady state, a "GIL hold" inferred from a lock-free endpoint that was not lock-free, and a vote-completion metric that scored the one captured 4918 ms stall as *fast*). A wrong-but-plausible reading here is the expensive failure mode, not a wrong edit.

### What is left to do

**One thing: capture a stall on a real labeling session, with the instruments on, and read off which phase expanded.** Everything else that could be done without the GRID has been done.

The original closing command below is **satisfied and must not be used** — the timing hook it asks for shipped in #3855, and a stall *was* captured with it on 2026-09-14. That capture is what opened the current line of investigation, not what closes the issue. The current criterion is:

> A full slate labeled on the merged build with no felt stall and no `detector_write` line over the bar — **or** a felt stall whose per-vote budget names the phase that expanded (see the reading table in [the 2026-09-16 comment](https://github.com/samggreenberg/VTSearch/issues/3853#issuecomment-5699088662)).

### How to run that session

```bash
export VTSEARCH_DIAGNOSE=1      # the whole bar set at once (#3875)
                                # do NOT also set VTSEARCH_GC_WARN_MS: it bypasses the coupling
vtsearch                        # launcher unchanged; it sets VTSEARCH_LOG_FILE itself

# second pane, so the filesystem and the process are on the record for the same minutes
python scripts/experiments/stall_3853/sample_host.py \
    --mount /exp/$USER --mount /expscratch/$USER \
    --match "python app.py" --out host-$(hostname -s).jsonl

# after
python scripts/experiments/stall_3853/analyze_app_log.py data/logs/app-*.log --votes 15
python scripts/experiments/stall_3853/sample_host.py --summarize host-*.jsonl
```

[...]

PR #3867, opened 2026-09-15
Move the per-vote detector-file rewrite off the request thread (#3853)

samggreenberg:
Closes the open item on #3853: capture the stall's cause and fix it.

## What the morning on the GRID established

The stall is **filesystem latency reaching the request path**, not a GIL hold. Full write-up in [the issue comment](https://github.com/samggreenberg/VTSearch/issues/3853#issuecomment-5680143087); the short form:

- The reviewer's own captured trace has eight vote POSTs over 250 ms. **Six were slow with nothing else in flight affected** — a GIL hold or GC pause cannot do that. The only thing the vote path does that nothing else does is `fsync` + rename the whole 650 KB detector JSON to `/exp`.
- The one event that froze every request (4918 ms) froze `jobs/active` too, which the issue read as "lock-free, so GIL". It is not lock-free: `get_context()` / `get_detector_context()` take `_state_lock` for every request. On the 09-14 build every request in the second after a vote also did a raw `stat()` of the detector file **inside** `_state_lock` (the window #3860 closed).
- `/exp` on rack7n06: cold 4 KB reads p50 5–7 ms, p99 41–67 ms, max 184 ms; the mount's WRITE RPCs average 75 ms exec over 82 days (63 ms queueing) against 0.4 ms for GETATTR; sibling `/exp` mounts on the node average up to 6.8 s per WRITE. `/expscratch`: 0.4 ms reads, 3.7 ms writes.
- A fresh process labeling one slate (600 votes, SigLIP semantic, learned sort every vote) never stalled: keypress→panel p50 84 ms / max 459 ms, zero watchdog events, zero major faults, every outlier a ~300 ms gen-2 GC pause.
- The watchdog did catch one `/exp` stall at dataset load: `cpu 0.02 of wall, majflt +245`, thread dump in `ctypes.CDLL` under `llvmlite` ← `numba` ← `import cuml` — a `dlopen` from the NFS venv with the GIL held.

## The fix

`vtscore/detectors/store.py` gains a background write queue. `sync_labels_to_loaded_detector` composes the merged labelset on the request thread (in-memory caches refreshed immediately) and hands the serialised text to the writer; the vote returns. Contract that keeps the file "true enough" while it lags:

- `_read_detector` returns the queued text, so `GET /api/detectors/<name>`, the rehydrate, export and every other in-process reader see a vote the moment it is accepted;
- `_write_detector` drains the queue for its path first, so a direct read-modify-write (the label routes, rename, `labelset-source` push) can never be overtaken by older queued text;
- a burst of votes coalesces into one write of the newest text; the in-flight entry stays visible to readers until it lands;
- unload flushes, both delete paths and rename discard (a queued write landing after an unlink would resurrect the file), `atexit` flushes;
- a failed background write is logged and raised as `DetectorWriteError` from the **next** vote, which the vote route already turns into the 500 it returned before;
- `VTSEARCH_DETECTOR_WRITE_MODE=sync` restores the inline write; the test suite runs that way (autouse in `tests/conftest.py`), and `tests/detectors/test_detector_write_queue.py` opts back into `async` to pin each clause above, including the vote route returning in under 2 s while every physical write is blocked.

## Kit fixes

- `drive_labeling.py`: percent-encode paths (every real detector name has spaces; all three per-vote `GET /api/detectors/<name>` had been failing client-side at 0 ms and reading as *fast*), and `--embedder-type` / `--text-query` so it can register the plain semantic detector the slates are actually labeled with.
- `analyze_app_log.py`: list slow phases and lock waits even with no `stall:` line — an `fsync` stall releases the GIL and never trips the watchdog.
- README: match the reviewer's shape; read the log even without a watchdog line.
- Lesson file: the deployed launcher was a pre-#3860 copy with no log file.

## Verification

- `ruff check` / `ruff format --check` clean on every changed file; pre-commit hooks passed on the GRID.
- `tests/detectors/test_detector_write_queue.py`: 8 passed. `tests/detectors`, `tests/api/{test_vote_provenance_api,test_api_contracts,test_error_recovery}.py`, `tests/core/{test_medias,test_media_list_edge_paths}.py`, `tests/datasets/test_multi_dataset.py`: 1308 passed (job 652356).
- Full suite on the final commit 7c65dd115 (job 652373, `suite.sbatch` on a separate worktree; the log's `=== HEAD` line names the commit): **11548 passed, 6 skipped, all gates green** in 4 min. A second run against the bare SHA (job 652376) is a confirmation only.
- Live before/after with a 5 s write stall injected at two votes, sync vs async: keypress→panel max **5104 ms → 326 ms**, vote POST max **5035 ms → 29 ms** (PR comment below).

## Ops notes (not in this PR)

- Put the data dir (`detectors_dir` at minimum) on `/expscratch`; consider the venv too.
- `~/.local/bin/vtsearch` on the GRID was re-copied from `scripts/slurm/vtsearch-slurm.sh` (backup `vtsearch.bak.20260915`) so the next allocation writes `data/logs/app-*.log`.


samggreenberg:
## Live before/after on the GRID: a 5 s write stall, injected

Two more instances from this branch on v100 nodes, identical except for `VTSEARCH_DETECTOR_WRITE_MODE`, each with the `knife` slate and the driver at 1 s think time for 420 votes. A wrapper (`run_app_inject.py`, in the scratch dir) makes the 200th and 350th physical detector-file writes block for 5 s before proceeding — `time.sleep` on the writing thread, which releases the GIL exactly as a blocked `fsync` does. That is the stall the reviewer feels, made to order.

| | `sync` (the old inline write) | `async` (this PR) |
|---|---|---|
| vote POST p50 / p99 / **max** | 28 / 42 / **5035 ms** | 21 / 29 / **29 ms** |
| keypress → panel p50 / p99 / **max** | 77 / 326 / **5104 ms** | 64 / 198 / **326 ms** |
| the two injected stalls, as the driver saw them | vote 198: POST 5029 ms, panel 5079 ms; vote 348: POST 5035 ms, panel 5104 ms | vote 198: POST 20 ms, panel 60 ms; vote 348: POST 21 ms, panel 61 ms |
| where the 5 s went | `process_request_thread` (the vote) | `detector-writer` thread; the file landed 5 s later |
| app log | `slow phase: label_sync total 5020ms (… queue=5011ms)` + `slow request: POST …/vote … 5026ms`, **no watchdog line** | nothing over the 300 ms bar (this instance predates the `detector_write` landing log added in 7c65dd115) |

In `sync` mode nothing else was blocked during the 5 s — the SPA's fan-out waits for the vote, and no poll happened to land — which is the "slow alone" shape six of the eight slow votes in the reviewer's trace had, and confirms the current `dev` no longer holds a lock across the write (the freeze-everything variant needs the pre-#3860 `stat()` under `_state_lock`).

## Same process, five slates, 2,400 votes: no growth

The first instance kept going after `knife`: `bowl`, `clock`, `cup` (with a polarity flip every 8th vote) and `chair` imported and labeled in the same process. 35 gen-2 GC pauses over the session, max 387 ms, no trend with datasets loaded or labels cast; zero lock waits, zero `labeling_status` slow phases from the flips, zero watchdog events, zero major faults.

## Verification

- Targeted run on the final commit (job 652370): `tests/detectors/test_detector_write_queue.py` 8 passed; `tests/detectors`, `tests/api/{test_vote_provenance_api,test_api_contracts,test_error_recovery}.py`, `tests/core/{test_medias,test_media_list_edge_paths}.py`, `tests/datasets/test_multi_dataset.py` 1308 passed.
- Full suite via `suite.sbatch` on a separate worktree: job 652373, result to follow.

samggreenberg:
Full suite on the final commit 7c65dd115 (job 652373, suite.sbatch on a separate worktree): **11548 passed, 6 skipped, 2 xpassed, all gates green** in 4 min.

samggreenberg:
Correction: job 652373 did **not** test this branch. I passed the local branch name, which is checked out in the dev worktree, so the checkout in the suite worktree failed and the run exercised dev (bfcd905b2). Resubmitted against commit 7c65dd115; real result to follow.

samggreenberg:
Retracting the correction: job 652373 **did** test this branch. Its log header reads `=== HEAD 7c65dd115 Log a slow queued detector write…` and `ref claude/stall-capture-3853 matches origin/claude/stall-capture-3853`; suite.sbatch restores the worktree's original checkout after the run, which is what I misread. So: 11548 passed, 6 skipped, all gates green on 7c65dd115. Job 652376 is a second run of the same commit, confirmation only.

Issue #3879, opened 2026-09-16
Below-cut probe covers only 6 of 25 classes: 50,024 unswept pairs, 1,900 labels to match the existing design


samggreenberg:
## The gap

The screened per-class pass (#3720) is complete — 25 classes, 12,632 candidates,
12,632 verdicts, zero delta (banked through #3878). Every one of those verdicts
is about an image the OWLv2 screen **flagged** for that class at or above its
cut.

The complementary question — *is the cut losing positives?* — has been asked for
only **6 of 25** classes: `bicycle`, `boat`, `dog`, `fire hydrant`, `sink`,
`stop sign`. Those are the `(any in image, no box)` probes banked as
`LABELSETS__belowcut__<class>.json`, 100 images each.

Nineteen classes have never had one.

## What the probe actually samples

Worth stating, because the name misleads: it is not "the detections below the
cut". Checking the banked images against `owl_queue.jsonl`, only **11–25 of each
100** carry *any* detection for their class — the rest were never flagged at all.
So the probe samples the class's whole unswept remainder, which is why the
question is "is there one anywhere in this image?" with no box to confirm.

The population per class is therefore `3,397 − known As − slate n`:

| | |
|---|---|
| images in the pass | 3,397 |
| unswept pairs per class | 2,183 (`bench`) – 3,236 (`bicycle`) |
| **19 classes with no probe** | **50,024 unswept (image, class) pairs** |
| coverage of a 100/class probe | **3.8%** |

## Cost

At the reviewer's measured pace — **19 labels/min**, averaged over the three
most recent sessions (car 588 in 33 min, umbrella 571 in 28, spoon 543 in 28):

| design | labels | time | coverage |
|---|---|---|---|
| match the existing probe, 100/class × 19 | **1,900** | **~1.7 h** | 3.8% |
| 300/class | 5,700 | ~5.0 h | 11% |
| exhaustive | 50,024 | ~44 h | 100% |

Build cost is negligible: the question carries no box, so it is 100 symlinks per
class plus 19 dataset/detector pairs — no rendering pass, unlike the boxed
slates.

## What it would buy, and why this is filed rather than run

**A bound, not a discovery.** The six probes already done came back
**0 good / 100 bad, all six — 600 for 600.** By the rule of three that puts the
95% ceiling on the per-probe miss rate at ~0.5%, i.e. ≤ ~14 missed positives in
a class's ~2,800 unswept pairs. Running 1,900 more and again finding nothing
[...]

Issue #3877, opened 2026-09-16
Move the VTSearch checkout off /exp: source, static/ and the diagnostic log still page in at 16ms (#3868 follow-up)


samggreenberg:
## What #3868 left behind

#3868 moved the venv (`/expscratch/sgreenberg/vtsearch-venv`) and the data dir
(`/expscratch/sgreenberg/vtsearch-data`) off `/exp`, on the grounds that `/exp`
page faults cost 5–180 ms. The **checkout itself** was not moved, so
`/exp/sgreenberg/projects/VTSearch` still holds the Python source the app
imports, `static/` (every frontend asset the browser fetches), and — since
#3860 — `VTSEARCH_LOG_FILE`, the diagnostic log.

## Measurement

Sampled with `scripts/experiments/stall_3853/sample_host.py` through the
2026-09-16 labeling session (543 vote cycles, 28 min, job 655193 on rack8n06).
Per-interval, differenced, never lifetime counters:

| mount | op | ops | exec p50 | max |
|---|---|---|---|---|
| `/expscratch/sgreenberg` | READ | 2,321 | **0.33 ms** | 0.77 ms |
| `/expscratch/sgreenberg` | WRITE | 2,041 | **0.77 ms** | 1.2 ms |
| `/exp/sgreenberg` | READ | 374 | **16 ms** | 20 ms |
| `/exp/sgreenberg` | WRITE | 60 | 2.0 ms | 3.0 ms |

`/exp` reads are **48x** slower than `/expscratch` reads. Over the same window
the app process took **+399 major faults** (RSS 1177 → 1761 MB) against those
374 `/exp` READs — near-identical counts, which is what you would expect if the
faults *are* the reads: the app paging its own code in from `/exp` at ~16 ms a
time.

## Why it is worth doing even though nothing hurt

It cost nothing measurable in that session — max server-busy time in any vote
cycle was 125 ms, and #3853 closed on it. So this is not urgent.

But it is the same mechanism as the one event #3853 ever captured with a thread
dump: on 2026-09-15 the watchdog caught a `dlopen` of llvmlite **from the venv
on `/exp`**, under the GIL, `cpu=0.02`, `majflt +245`. That venv is now on
flash; the source tree is not. A cold import on the request thread still costs
16 ms per page here and 0.33 ms one mount over.

There is also a self-inflicted bit: the diagnostic log for a latency
investigation is being written to the slow mount, which adds `/exp` WRITEs on
the logging path at INFO.

[...]

PR #3882, opened 2026-09-16
Build the below-cut probe for the nineteen classes that never had one (#3879)

samggreenberg:
Adds `make_belowcut_slates.py` and builds the 19 probes. Dashboard is loaded: **19 datasets + 19 detectors, 100 images each, 1,900 to review.**

## Why this script exists

The six probes already run (`bicycle`, `boat`, `dog`, `fire hydrant`, `sink`, `stop sign`) were built by a script that was **never committed** — only `belowcut.json` survived. So the population rule had to be reverse-engineered from that manifest. It is:

```
every image in the pass - the class known As - the class own slate - anything already answered
```

Five of the six reproduce exactly under it. `bicycle` is 72 off in a way I could not reconstruct; it is already probed, so it is not rebuilt here and the discrepancy is recorded rather than papered over.

## The name is misleading, and it cost me a 10x error

"Below cut" reads as *the detections that scored under the threshold*. It is not. Of the 100 images in each existing probe, only **11-25 carry any detection for their class at all** — the rest were never flagged. The population is the class whole unswept remainder, **~2,000-2,900 per class**, not the few hundred sub-threshold detections. I costed this issue wrong the first time by reading the name instead of the data, and the docstring now says so up front.

## What was built

| | |
|---|---|
| classes | 19 (the 6 already probed are skipped) |
| images per class | 100, matching the existing probes exactly |
| total | **1,900** |
| seed | 3879, fixed, recorded in the manifest |
| broken symlinks | 0 |

Verified independently against the banked slate labelsets: **zero overlap** with any class own slate verdicts or its designated positives. 428 of 1,900 picks are images another class also sampled — expected, since each class samples its own remainder independently, and the existing six have the same property.

Detector names reproduce the existing convention exactly (`review_name(cls)` + ` (any in image, no box)`), verified against all six banked probes. Every name leads with the class, so the class is on screen while voting.

Imported by job 655345 (GPU, 2s per slate) into `/expscratch/sgreenberg/vtsearch-data`. The job excluded rack8n06 because a parallel session is taking host-wide NFS measurements there for #3877.

## Note for #3877

`import_slates.py --data-dir` still defaults to the hardcoded `/exp/sgreenberg/projects/VTSearch/data`, and `bank_verdicts.py` hardcodes `REPO = /exp/sgreenberg/projects/vts-annq-3720/...`. Both are `/exp` paths baked into committed scripts. Passed explicitly here; flagged for that issue.


samggreenberg:
## Correction: the claim that no committed builder existed is false

This PR's description says the six earlier probes "were built by a script that
was **never committed**", and that the population rule "had to be
reverse-engineered from the manifest it left behind". Both statements are wrong.

`scripts/experiments/pile/make_belowcut.py` has been on `dev` since
**2026-09-08** (#3768), carrying the entire design:

```python
FINISHED = ["bicycle", "boat", "dog", "fire hydrant", "sink", "stop sign"]
SAMPLE = 100
SEED = 3768
```

`load_belowcut.py` sits beside it to import the result. The correct change for
#3879 was a class list, not a second builder.

**How the error happened:** I grepped `/expscratch/.../vlm-3720/*.py` — the
scratch directory where this study's ad-hoc scripts live — found nothing, and
concluded it had never been committed. I never ran `git grep belowcut` against
the repo, which finds it immediately.

**What is unaffected:** the two population rules are equivalent. Mine was
`all images − known As − slate − answered`; the original is `top score < cut,
excluding known As and answered`. An image is in a slate iff it cleared the cut
and was neither known nor answered, so the complements coincide — which is
exactly why five of the six reproduced under the reverse-engineered rule. The
1,900 verdicts and 13 positives banked under #3879 are sound.

**What it did cost:** the `bicycle` discrepancy I could not reconcile (2,487
recorded vs 2,415 computed) is explained — the original sources `answered` from
`human_record` globs over `LABELSETS__slate__*` and `LABELSETS__pass25__*`, mine
from `banked_labels.json` via a two-class map. `bench` is the other class with a
`pass25` file, so its probe sampled a slightly different population than the
established design. The verdicts stand; the deviation is worth knowing.

Fixed in #3898: the duplicate is deleted and `make_belowcut.py` gains
`--classes` and `--seed`.

---

Issue #3903, opened 2026-09-16
DocMarks: build_corpus should refuse to re-tier an existing corpus without --pin-tiers


samggreenberg:
**Difficulty: small.** One guard and its tests.

**Recommended Claude model: Sonnet 5.**

Carried out of #3343 so it survives that issue closing.

## The trap

`build_corpus.py --pin-tiers <build_report.json>` keeps tier membership stable across a later build over a different page set. Without it, `assign_tiers` recomputes cutoffs from the new page set, so pages silently move between `s` / `m` / `l` — and every number measured on the v3 cells (tiers `s` and `m`, and tier `l` now building) stops being comparable to the next build's, with nothing reporting it.

Today pinning is only **advice**: `GRID-RUNBOOK.md` "Growing the corpus later" and a comment in `build_report.json`. It is exactly the kind of mistake a paragraph does not prevent.

## Fix

When `--out` already holds a `build_report.json` with `tier_cutoffs`, refuse to build unless the caller passes either `--pin-tiers <that report>` or an explicit `--new-version` (which says "this is a new, incomparable corpus version" and records it in the new report). Fail before any pull or clustering starts, and name both flags in the message.

## Tests

- existing report + neither flag → refused before work starts;
- `--pin-tiers` → cutoffs reused, tiers unchanged for pages in both builds;
- `--new-version` → allowed and recorded;
- empty `--out` → unchanged behaviour.

samggreenberg:
Addressed in #3906

samggreenberg:
Shipped to `main` in the 2026-09-17 release — fixed in #3906.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #3905, opened 2026-09-16
GRID suite blocked by pip-audit: setuptools 81 (PYSEC-2026-3447) and install.sh pins <82


samggreenberg:
**Difficulty: small.** One version pin plus an upgrade of the shared venv, but the upgrade is shared state.

**Recommended Claude model: Sonnet 5.**

## What broke

`./run-tests.sh` on the GRID now ends in **RUN FAILED: pip-audit** even when every test passes. Seen on suite job 655642 (2026-09-16, `dev` + #3886): **11,659 passed**, pyright/vulture/npm audit/Vitest OK, then:

```
setuptools 81.0.0  PYSEC-2026-3447  83.0.0
TESTS BLOCKED: pip-audit failed
```

The advisory is new and has nothing to do with the code under test, so every GRID suite run fails this gate until the venv changes.

## Why it won't fix itself

`scripts/install.sh` pins `pip install --upgrade pip "setuptools<82" wheel` (since 76813f0ea, 2026-06-05, "to silence pip conflict error"). The fixed version is **83.0.0**, so reinstalling or rebuilding the venv lands back on the vulnerable version.

## Done when

- Find out which dependency conflicted with setuptools ≥ 82 in June, and whether it still does (setuptools 82 removed `pkg_resources`, the likely reason).
- Raise the pin to `>=83` (or ignore this one advisory in pip-audit, if something still needs `pkg_resources`, with the reason recorded).
- Upgrade the shared GRID venv (`/expscratch/sgreenberg/vtsearch-venv`, and the `/exp` checkout's `.venv` the suite uses). Other sessions' jobs share it, so tell them first.
- A GRID `./run-tests.sh` with pip-audit green.

Needs the GRID: the shared venv only exists there.

samggreenberg:
Code half in #3909: `install.sh` now picks setuptools from the installed torch (`>=83.0.0`, or torch's own cap for torch 2.11–2.12.x) instead of a flat `<82`. Root cause: 76813f0ea pinned `<82` for torch 2.12.0, and the venv rebuilt on 2026-09-15 for #3877 got 81.0.0 because of it.

### Upgrade the shared venv — after tier `l` finishes, not before

[...]

PR #3906, opened 2026-09-16
DocMarks #3903: refuse to re-tier an existing corpus without --pin-tiers

samggreenberg:
Closes #3903.

`assign_tiers` recomputes tier cutoffs from whatever page set it is given. So a build pointed at an `--out` that already held a finished build moved pages between `s` / `m` / `l`, and every cell built on the old tiers silently stopped being comparable. The only thing preventing it was a paragraph in `GRID-RUNBOOK.md`.

## What changed

**`build_corpus.py` decides before the pull** (`tier_provenance`). If `--out` already holds `build_report.json`, the build has to say which tier promise it makes:

- `--pin-tiers <build_report.json>`: same corpus, tiers held by the recorded cutoffs;
- `--new-version`: a new, incomparable corpus version;
- neither: refused, exit 2, with a message naming both flags. Nothing is written into `--out`.

It also refuses up front:
- a `--pin-tiers` file with no `tier_cutoffs`, which used to surface only after clustering;
- both flags together;
- an unreadable report. Something was built there, so it isn't treated as an empty directory.

The pin file is now read at the start, which matters because the natural argument, `$OUT/build_report.json`, is the file the build overwrites at the end.

**The report records the choice** under `tier_provenance`: `pinned_from`, `pinned_cutoffs`, `new_version`, and, for a new version, the `superseded_cutoffs`. The break is then on disk.

**`launch_docmarks.sh build` refuses before submitting** unless `VTS_DOCMARKS_PIN_TIERS` or `VTS_DOCMARKS_NEW_VERSION=1` is set, so the refusal doesn't wait in the queue first. Checked by hand against a throwaway `--out` holding a report: `REFUSED`, rc 2, nothing submitted.

**Docs.** The stage-3 roster rebuild (runbook and README quickstart) now passes `--new-version`. It builds into the candidate build's `--out`, and picking a roster changes which pages are positives, so its tiers do move. That's harmless before any cell exists, and this flag records it. "Growing the corpus later" now says the rule is enforced.

## Tests

9 new, in `TestRebuildingAnExistingCorpusMustChooseATierPromise`, with `load_anchor_sources` rigged to raise so "refused before the pull" is asserted, not assumed:

- unpinned build into a finished corpus → exit 2, pull never reached, `--out` untouched;
- the refusal names `--pin-tiers`, `--new-version` and the report;
- `--pin-tiers` → recorded cutoffs handed through, guard passed;
- rebuilding the same page set with round-tripped cutoffs gives page-for-page identical tiers;
- `--new-version` → allowed, superseded cutoffs recorded;
- empty `--out` → unchanged;
- unreadable report, pin file without cutoffs, both flags → refused.

`tests_lib/datasets/test_docmarks_corpus.py`: 322 passed on a GRID cpu node. The full suite is running; the result will be posted below.

Written with no access to the live corpus: tier `l` cells are building from `/expscratch/sgreenberg/docmarks/corpus`, and nothing here touches it.


samggreenberg:
**GRID suite, job 655678 @ `ac1804a3e`:** **11,668 passed**, 6 skipped, 2 xpassed. pyright, vulture, npm audit and Vitest are OK.

The run still ends **RUN FAILED**, but only on `pip-audit`: `setuptools 81.0.0 PYSEC-2026-3447` in the shared GRID Python environment. That's #3905 and has nothing to do with this change; the same failure hit the #3886 suite (job 655642).

The 2 xfail tests reported as xpassed this run where they xfailed on 655642. Neither is in the DocMarks tests.

---

Issue #3814, opened 2026-09-12
A corrections row records no rule version, so a ruling silently changes what old rows mean


samggreenberg:
**Difficulty: low.** One optional field, the writers that set it, and a reader that can ask the question.

**Recommended Claude model: Sonnet 5.** Mechanical once the decision below is taken.

## Why

A row in `corrections.json` records *what* a reviewer answered and never *which
question they were shown*. The class rule is the question — it is the detector
name the reviewer reads while voting (#3612) — so when a ruling changes a rule,
every row cast under the old one silently starts meaning something else.

That is not hypothetical: it cost the 80-image planter recheck. `vase` claimed
"flower pots, planters" until #3784 withdrew the line, and nothing on the 31
`vase` rows or the 49 `bowl` rows said which wording they were cast under. The
only way to find out was to ask a human all 80 again (#3778), of which **21**
turned out to have moved and 59 had not. Re-asking is the right move when the
answer is unknown; the point here is that it was unknowable from the file.

Three rulings have now landed on classes with rows already on disk — planter
(#3784), dead birds (#3791), fork tines (#3792) — and 13 of the 25 classes have
not started their pass, so more are coming.

## Proposed

One optional field on a correction row:

```json
{"image_id": 918, "class": "vase", "present": false, "rule": "vase not planters", ...}
```

written by whatever creates the row (`verdicts_to_corrections.py`,
`ingest_slate.py`, `apply_recheck.py`) from
`SCALE_CLASS_RULES[cls].name` at the time. Then "which rows predate the ruling
now in force?" is a query rather than an 80-image review, and a recheck slate
builds itself.

Optional, because the 872 rows already on disk cannot be back-filled honestly —
a row with no `rule` means *unknown*, and it must not be silently read as the
current one. `apply_recheck.py` already checks the labelset's recorded rule
against the one in force and refuses when they differ; this is the same check
made available to the rows themselves.

## To rule on rather than assume

1. **Does an unchanged rule name mean an unchanged rule?** The name is the
   wording the reviewer saw, so it is the right key for that question, but a
   rule can be edited in the body without the name moving (#3756 changed
   `bench`'s Bad list and kept its name). A content hash answers the stricter
   question and is unreadable in a diff; the name is readable and coarser.
   Recording both is cheap.
2. **Should a row whose rule has moved be reported by the build**, or only by a
   tool that is asked? A build that prints "N rows predate the rule in force"
[...]

PR #3916, opened 2026-09-17
Re-vote dog and fork under the rules in force; a re-vote disagrees ~4% on its own

samggreenberg:
Re-vote dog and fork under the rules in force; a re-vote disagrees ~4% on its own

dog    59 good /  14 bad =  73   rule 'dog'               -> 'dog not wolves'
fork  130 good / 266 bad = 396   rule 'fork incl plastic' -> 'fork incl sporks not strainers'

Both complete (uniq == slate, 0 foreign), both stamped with the CURRENT rule and
rule_digest. This replaces slate labelsets voted under superseded wording, which
#3814's stamp had kept out of silence_rate and out of the corrections
application. Same pre-boxed images as the pass; only the rule name shown changed.

This is a supersession, not a loss: every image was re-voted, so no verdict cast
under the old wording survives unstamped. Git keeps the old files.

WHAT THE RE-VOTE MEASURED

Image by image, old vote against new:

  dog    agree 69/73   good->bad 2   bad->good 2
  fork   agree 382/396 good->bad 9   bad->good 5

`dog` is a natural control. Its rule change is a NARROWING ("not wolves"), which
can only turn a Good into a Bad -- yet it flipped two in each direction. The two
bad->good flips cannot be the rule, and the symmetry suggests all four are simply
re-vote inconsistency.

`fork` then moved LESS than that:

  dog  disagree   4/73   5.5%  [1.5%, 13.4%]
  fork disagree  14/396  3.5%  [1.9%,  5.9%]
  pooled         18/469  3.8%  [2.3%,  6.0%]

The intervals overlap heavily. So neither rule change produced a difference
distinguishable from re-voting noise at this sample size, and re-voting the same
pre-boxed image weeks later disagrees about 4% of the time.

That second number matters beyond these two classes: it is a direct estimate of
same-reviewer disagreement on box confirmation, and every banked verdict in this
pass carries it. Caveat: `dog`'s wording did change, so this is near- but not
purely test-retest.

fork carries 3 reviewer-redrawn Goods (2348050, 285639, 2358315). 2348050 is the
case where the stored box sat on a plastic KNIFE while a real fork was elsewhere
in the frame -- redrawn so the correction points at the fork, not the knife.

vase (394) is still to be re-voted.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


samggreenberg:
## Correction: the ~4% is concentrated on ambiguous images, not spread across every verdict

This PR's description says re-voting disagrees ~4%, and that "every banked
verdict in this pass carries it." The second half overstates it.

The reviewer confirmed there were **no wolves** in the `dog` slate, so the rule
change ("dog" → "dog not wolves") could not have moved a single vote — `dog` is a
clean control, now confirmed directly rather than inferred from the flips being
symmetric. And the four flips were, in the reviewer's words, **fluffy animals
that were a blurry mess**, where the judgment call went one way or the other.

That changes what the number means. It is not a uniform error rate that any
verdict might carry. It is a measure of how many images sit on a **genuine
judgment boundary** — ambiguous enough that the same reviewer, re-voting, can
reasonably land either side, with neither answer wrong. Clear cases do not flip.

The practical difference: a uniform model would discount every verdict slightly;
the correct reading is that a small, identifiable subset of images is inherently
uncertain, and those are what to flag. Whether `fork`'s 14 flips share this
mechanism is plausible but was not observed.

---

Issue #3992, opened 2026-09-18
The lumping defect is per-image, not per-class: every median in C is ~1.00 and 'clean' classes carry an 8-18% pile tail


samggreenberg:
**Difficulty: medium-hard.** The measurement exists; what is missing is a signal COCO can produce on its own.

## Why a class-level verdict is the wrong instrument

#3985 named four classes in *C* whose COCO boxes are drawn around groups — `book`, `car`, `kite`, `boat` — on a **mean** area ratio against LVIS. Re-running the same measurement with a median and a per-image rate (`scripts/experiments/pile/coco_box_granularity.py`, extended to all 25 classes of *C*) says that mean is a tail:

| class | mean area ratio | **median** | **% of images ≥1.6** |
|---|---:|---:|---:|
| `book` | 1.83 | 1.22 | **31%** |
| `boat` | 1.76 | 1.01 | 18% |
| `sink` | 1.42 | 1.01 | **18%** |
| `kite` | 2.08 | 1.01 | 16% |
| `bowl`, `truck` | 1.15, 1.18 | 1.01 | 12% |
| `cup` | 1.15 | 1.02 | 10% |
| `bench` | 1.36 | 1.00 | 10% |
| `bird`, `chair`, `bottle` | 1.32, 1.16, 1.44 | ~1.01 | 8% |
| `car` | **2.09** | 0.99 | **7%** |
| `clock`, `cell phone`, `bus`, `dog` | 1.03–1.07 | ~1.00 | 3–5% |

Two readings follow:

1. **Every median is ~1.00 except `book`.** A class called a lumper is one whose *minority* of images carry a pile — `boat`'s 1.76 is a handful of marinas (LVIS's other names on those images: `mast` 586, `buoy` 295, `sail` 145), and `car`'s 2.09 rests on 7% of its images.
2. **The per-image rate does not split *C*.** `sink` at 18% matches `boat`; `bowl` and `truck` at 12% beat `car`'s 7%. So #3985's **option 2 (drop the four classes) does not leave a clean *C***, and it pays two scene-exclusive anchors (#3603) for a fraction of the error.

The defect is per image. The guard should be too.

## What is actually needed

`band_for` rejects an image when the union outgrows the largest box (scatter). The mirror test — *is this box around a group?* — has no signal yet, and it must come **from COCO alone**: LVIS is a selection instrument for measuring this, not a build dependency, and it covers as little as 16% of a class's COCO images.

[...]

Issue #3985, opened 2026-09-18
The scatter guard cannot see a stacked pile: COCO's `book` boxes are 1.8x LVIS's for the same books


samggreenberg:
**Difficulty: medium-hard.** The defect is measured and undisputed; what to *do* about it is a design decision that moves band membership.

**Recommended Claude model: Opus 4.8.** Any fix changes which images land in which band, which is the thing every scale result is conditioned on — a wrong-but-plausible rule would silently restate #3156.

## The gap

`band_for` rejects an image when a class's boxes are **spread out**: the union outgrows the largest box by more than `BAND_MAX_INFLATION`, so the union describes the scatter rather than an object.

It cannot catch the opposite failure, and that one is harder. **A stack is compact.** Five books on a shelf boxed as a single lump have a union equal to that lump, an inflation ratio of **1.0**, and they pass the guard cleanly. The image is then banded by the size of the *pile*, and nothing downstream can tell — exactly the class of error `docs/experiments/2026-09-17-vg-scale-corrections/REPORT.md` calls "a quieter error than a wrong label, and one nothing downstream can detect".

COCO's explicit mechanism for a group is `iscrowd=1`, and **it is not the vector**: only 2.6% of `book` annotations carry it (`kite` 3.0%, `bird` 2.5%, most of *C* under 1%). The lumping happens inside ordinary `iscrowd=0` boxes.

## Measured

`scripts/experiments/pile/coco_box_granularity.py`, against LVIS on the same images. Instance counts alone cannot answer it — COCO drawing one box round a pile (**lumping**) and COCO annotating fewer instances of the same size (**incompleteness**) both raise LVIS's count. Box **area** separates them: under lumping COCO's mean box is much larger for the same object; under incompleteness they match.

| class | imgs | count ratio | COCO box | LVIS box | **area ratio** | verdict |
|---|---:|---:|---:|---:|---:|---|
| `banana` | 331 | 7.58 | 13.0% | 7.4% | **8.15** | lumps |
| `orange` | 155 | 4.62 | 8.0% | 4.4% | **6.77** | lumps |
| `apple` | 166 | 4.48 | 7.2% | 3.8% | **5.69** | lumps |
| `car` | 355 | 1.40 | 4.6% | 5.7% | **2.09** | lumps |
| `kite` | 334 | 1.20 | 6.2% | 6.5% | **2.08** | lumps |
[...]

PR #3993, opened 2026-09-18
Measure all of C for lumping, and report the median beside the mean (#3985)

samggreenberg:
Closes nothing on its own; it corrects the evidence #3985 was filed on, and files #3992.

Sam read the four-class list in #3985 and asked why `boat` was in it when `bird` should be commoner and cups stack. Checking that turned up two gaps, neither visible in the script's output.

## Ten classes of *C* were never measured

`SAME` carried 21 entries; *C* has 25. `backpack`, `bench`, `bicycle`, `cell phone`, `fire hydrant`, `fork`, `sink`, `spoon`, `stop sign` and `vase` had no reading at all — which reads as "clean" in a table and means "never asked". LVIS names all ten 1:1. Measured:

- **`sink` 1.42, with 18% of its images over the threshold** — `boat`'s rate;
- `bench` 1.36; `fork` 1.12 and `spoon` 1.13 (worth checking — cutlery in a drawer is a pile by construction);
- `fire hydrant` has no LVIS overlap to read, and the script now **says so** rather than dropping the class silently. A class with no reading and a class with a clean reading are not the same claim.

A related trim: the issue's table printed 13 of the 21 rows the script produced, omitting seven classes of *C* (`bird` 1.32, `chair` 1.16, `bowl`/`cup` 1.15, `umbrella` 1.07, `knife` 1.06, `bus` 1.05) while keeping five non-*C* reference classes. The "nothing in between" claim survives — `bird` at 1.32 still sits under `bottle`'s 1.44 — but it was argued from a third of the measurement.

## The class mean is a tail statistic

Every median in *C* is ~1.00 — `boat` 1.01, `car` 0.99, `kite` 1.01 — and only `book` moves at 1.22. So "COCO lumps `boat`" states that a minority of its images carry a pile (18%, the marinas), not that its boxes are typically around groups; `car`'s 2.09 rests on 7% of its images.

The per-image rate does not split *C* the way the means do — `sink` 18%, `bowl` and `truck` 12%, `cup` 10% against `car`'s 7% — so **dropping the four named classes would not leave a clean *C***. That answers #3985's option 2 and argues for option 4, filed as **#3992**.

## A docstring that contradicted its own table

The docstring said a class whose LVIS counterpart splits into subtypes is "left out rather than guessed at" and named `bottle` as the example, while `bottle`, `bird` and `cup` all sat in `SAME`. Keeping them is right — a floor on a class in *C* beats no reading — so the prose is corrected and `SUBTYPE_FLOOR` now marks them in the output: LVIS spells 676 of the birds on COCO's `bird` images `pigeon`/`gull`/`flamingo`/`duck` against 1,883 plain `bird`, so `bird` 1.32 and `cup` 1.15 are lower bounds.

## Changes

- `coco_box_granularity.py`: `SAME` extended to all 25 classes of *C*; `SUBTYPE_FLOOR` added; median, per-image `over` share and `cover` columns printed and written to `--out`; classes under the floor reported as `NOT COMPARED` with their COCO image count; docstring corrected on both points.
- `docs/experiments/2026-09-18-coco-only-supply-3983/REPORT.md`: the full 24-class table, the median-beside-the-mean section, the coverage and subtype-floor caveats.

## Verification

Re-run on the same sources (`vts-cache/coco_anchor` + `lvis_v1_val.json`, 19,516 shared images) reproduces every number in the shipped table exactly. Eight pre-pytest gates green. Suite: 11,832 passed, 58 skipped, 1 failed — `test_path_traversal_relative_rejected`, which fails at `HEAD~1` and at `fbbcada9a` checked out at this same path, so it is the worktree depth and not this branch (#3995).


samggreenberg:
**Suite: 11,832 passed, 58 skipped, 1 failed** (job 660810, 4:01) — and the one failure is not this branch.

`tests/io/test_label_import_endpoint.py::...::test_path_traversal_relative_rejected` fails at `HEAD~1` (= `origin/dev` at `f73d70254`) in this worktree, and also at `fbbcada9a` checked out **at this same path**, while passing at `fbbcada9a` in `projects/VTSearch` — same node, same job. The variable is the checkout's directory depth, not the commit: `../../../etc/shadow` resolves to the real `/etc/shadow` from a three-deep path, and the importer opens it (500) instead of rejecting the traversal (400), because no confinement is implemented. Filed as #3995.

Eight pre-pytest gates green (`check-docs`, `check-calibration-index`, `deptry` and the rest).

Issue #3997, opened 2026-09-18
Settle the VG diversity claim before the coco_quarry rebuild: off-COCO is differently composed at 0.53-0.55 matched, and the shipped probe split on the wrong flag


samggreenberg:
**Difficulty: medium.** The composition half is done (below). The hardness half is a detector run and a matched control.

**Recommended Claude model: Opus 4.8.** The trap is the label-quality confound, which is invisible in the numbers and inverts the conclusion.

The `coco_quarry` plan ([`docs/plans/coco-quarry.md`](../blob/dev/docs/plans/coco-quarry.md)) names this as the item that runs **first**, because it is the only live argument for keeping Visual Genome and the only one that could still change the migration decision: `anchor_to_coco` claims dropping VG's non-COCO half loses "VG's non-COCO diversity for nothing", and that claim has never been measured. It had no issue; this is it.

## First: the shipped probe splits on the wrong flag

`scripts/experiments/pile/provenance_probe.py` predicts `labels_exhaustive` and reads the result as provenance. But the loader defines those two flags as different claims, and says so:

> `labels_exhaustive` is also set by a human looking at ONE class; this says COCO answered for all eighty at once

So `labels_exhaustive` is true for *a COCO image **or** an image a human reviewed*. Measured on today's cell, that puts **2,138 reviewed-but-off-COCO positives on COCO's side — 41% of all 5,237 positives**. The gap has widened since #3670 ran the probe, because #3926's annotation pass folded 4,709 corrections into `exhaustive`. `coco_scored` is the flag that means "in COCO".

| cell | positives | `labels_exhaustive` | `coco_scored` | off-COCO |
|---|---:|---:|---:|---:|
| `vg_scale__siglip` | 5,237 | 4,488 (85.7%) | **2,350 (44.9%)** | **2,887 (55.1%)** |
| negative pool | 10,900 | 10,900 (100%) | 10,900 (100%) | 0 |

Re-running the probe on `coco_scored` (`siglip` / `siglip2_l` / `clip`):

| partition | siglip | siglip2_l | clip |
|---|---:|---:|---:|
| shipped flag, whole cell | 0.774 | 0.787 | 0.780 |
| `coco_scored`, whole cell | **0.850** | **0.862** | **0.845** |
| `coco_scored`, positives only | 0.577 | 0.580 | 0.575 |

[...]

Issue #3926, opened 2026-09-17
Apply the per-class pass to corrections.json: 3,838 positives validated in a dry run, gated on a small-band risk in four classes


samggreenberg:
## The per-class pass is banked but not in the benchmark

`vg_scale` is built from `corrections.json` alone — the pile build never reads the banked labelsets. That file has **872 rows and was last written on 2026-09-12**. Since then the pass banked **3,859 confirmed positives** (12,632 verdicts across 25 classes) and the below-cut probe (#3879) found **13 more**. None of it is in the benchmark.

This issue records a validated dry run of applying it, an audit of the one real risk, and the decision that gates doing it for real. **Nothing has been written.**

## Dry run: validated, then applied to scratch

The harness first had to reproduce the live file, because the converter rebuilds the whole file from its inputs:

```
verdicts_to_corrections.py (all defaults, --out scratch) + apply_recheck.py
  -> 872 rows vs live 872, 0 differences, identical source counts
```

With the pass's verdicts added (rule stamps carried, never re-derived — #3814), for all 25 classes:

| | rows | added |
|---|---|---|
| **A — the pipeline exactly as it stands** | 872 → 4,710 | **3,819 exclusions** + 19 boxed positives |
| **B — attach the stored OWLv2 box** | 872 → 4,710 | **3,838 new boxed positives** |

Both drop 0 rows. **Variant A is what "just run it" does, and it wastes the pass**: 3,856 of the 3,859 slate Goods carry no reviewer box, because the pre-boxed slate made each verdict a single click, and a boxless human "present" becomes an *exclusion* in the converter, not a positive. The box the reviewer confirmed lives only in `slates.json` — all 3,859 are there, so **B needs no OWLv2 rerun**. Consistent with the owner's ruling in `make_class_slates.py`: "a loose box costs a stratum, not a label."

## Three traps for whoever runs it for real

1. **`verdicts_to_corrections.py --out` defaults to the live file, and there is no `--dry-run`.** It does refuse to *drop* rows without `--allow-loss`, but a run that only *adds* rows passes that guard and overwrites the live file.
2. **Stopping after the converter silently undoes the planter ruling (#3784).** 21 retired planters flip back to positives. `--allow-loss` does not catch it — the rows *change*, they are not dropped. `apply_recheck.py` must run afterwards, every time.
3. **`apply_recheck.py --corrections` also defaults to the live file (or `$VTS_CORRECTIONS`) and writes in place.**

[...]

PR #3998, opened 2026-09-18
Settle the VG diversity claim: off-COCO is differently distributed, not harder (#3997)

samggreenberg:
Answers #3997, the item the [`coco_quarry` plan](../blob/dev/docs/plans/coco-quarry.md) says runs **first** — `anchor_to_coco`'s claim that dropping VG's non-COCO half loses "VG's non-COCO diversity for nothing" is the only live argument for keeping Visual Genome, and it had never been measured. It also had no issue until today.

**Differently distributed, but not harder. The difficulty argument does not survive.**

## The instrument was reading the wrong partition

`provenance_probe.py` predicted `labels_exhaustive` and called it "COCO-sourced". The loader sets two flags and documents them as different claims where it sets them — `labels_exhaustive` is true for a COCO image **or** an image a human reviewed for one class, while `coco_scored` means COCO answered for all eighty at once. On today's cell the loose flag puts **2,138 reviewed-but-off-COCO positives on COCO's side, 41% of all 5,237 positives**; the gap widened when #3926 folded 4,709 corrections into `exhaustive`, so the number was defensible when #3670 took it and is not now.

Two more corrections follow from the same table. The **pool must be excluded** — it is 100% `coco_scored` against positives that are 55% off-COCO, so a whole-cell probe scores positives-vs-negatives under a provenance heading (0.85, meaningless). And **composition must be matched away** — off-COCO supply runs 4 to 82 per cell, so class and band are themselves readable as provenance.

| | pooled | matched within `class@band` |
|---|---:|---:|
| siglip | 0.571 | **0.542** |
| siglip2_l | 0.578 | **0.539** |
| clip | 0.568 | **0.536** |
| clip_l | 0.576 | **0.532** |
| dinov3_patch | 0.563 | **0.526** |

About half the readable signal was composition. Per-cell fits are **not** usable at this supply (35 cells clear 25/arm; their AUCs run 0.22–0.75, straddling chance both ways), and the report says so rather than offering a per-class ranking.

## Readable is not harder, and the statistic decides what is visible

`provenance_hardness.py` fits the **shipped** head (`vtscore.training.svm.fit_linear_svm_head`) per cell and ranks held-out positives against a held-out half of the pool. The obvious statistic — the fraction of the pool a positive outranks — **saturates** at 0.85–1.00 and reads as a flat nothing (+0.000 to +0.006). The same fits re-aggregated on statistics that don't saturate lean the other way:

| statistic (+ = off-COCO ranks better) | siglip | siglip2_l | clip |
|---|---:|---:|---:|
| log1p(rank above) | +0.11 ± 0.06 | +0.15 ± 0.06 | +0.15 ± 0.07 |
| share in the pool's top 1% | +0.018 ± 0.013 | +0.027 ± 0.012 | +0.026 ± 0.016 |
| cells agreeing in sign | 35/58 | 37/58 | 36/58 |

Per-image ranks are stored, so `--aggregate` re-reads them under a different statistic **without refitting** — which is how the ceiling was found after the fits had already run.

## The control brought its own confound, and reporting it alone would have doubled the result

The verified-only arm (restricted to (image, class) pairs #3926 confirmed) is twice the effect: +0.013–0.017 on percentile, +0.25–0.31 on log1p. It is **not** the clean number. #3926 verified by asking whether a human could **see** the object, which selects for clear instances, while the COCO arm keeps every small and occluded instance COCO annotates. It bounds how much of the gap is label quality — real, given VG's 0.61 recall and 8.3% wrong-instance boxes (#3924, #3925) — but it does not reveal true difficulty. The unrestricted arm answers the plan's question.

## Changes

- `provenance_probe.py` — splits on `coco_scored`, excludes the pool, adds the matched-within-cell draw, and records all three corrections in its docstring.
- `provenance_hardness.py` (new) — the hardness measurement plus a no-refit `--aggregate` mode.
- `figures_3997.py` (new) + two figures — the matched-vs-pooled bars, and the per-cell differences as a **distribution**, so the mean is not read as a separation.
- `docs/experiments/2026-09-18-vg-diversity-3997/` — report and measurements, indexed in `docs/experiments/README.md`.

## Verification

Suite green: **11,842 passed**, 58 skipped, 1 xfailed. Eight pre-pytest gates and `ruff` clean. An earlier run failed 6 tests in `tests_lib/meta/test_docs_gate.py` because a new study directory needs a row in the experiments index — that gate is separate from `check-docs.py` and not among the eight, which is worth knowing for any commit that adds a directory there.

Both scripts reproduce their numbers exactly across three independent runs.


samggreenberg:
Merged to `dev` as 8570a80152a8d85d12aa51f5eaa0196022eb79b2 at 17:01:46 -0400. The `gh pr merge --merge` call returned a 502 from GitHub, but the merge itself had already gone through server-side — `c5eed6706` and `96e6e6cee` are both ancestors of `origin/dev` — and GitHub's record of this PR did not catch up. Closing it by hand so it does not sit open against work that has landed; the commits are in `dev`, nothing needs re-merging.

Issue #4003, opened 2026-09-18
corrections.json can no longer be regenerated: 6 of 8 recipe inputs were deleted, 4 with no copy


samggreenberg:
**Difficulty: medium. Urgency: high** — not because anything is broken today, but because every week that passes makes the four lost inputs harder to reconstruct from memory.

## The live file is intact. Its recipe is not.

`corrections.json` is fine: **4,709 rows**, sha256 `6fea3a0c257e175086d68eafb03db3c8`, now also copied to `/expscratch/sgreenberg/keep/corrections-backup-20260918/` with the copy hash-verified. No human judgement has been lost from the file.

What is gone is the ability to **regenerate or audit** it. `verdicts_to_corrections.py` rebuilds it from a declared recipe, and a housekeeping pass on 2026-09-18 (#4001) deleted the study directories that recipe reads from. Checked against disk:

| default input | status |
|---|---|
| `vgscale-3156/verdicts_20260820b.json` | **missing** — recoverable, a copy is in `keep/human-artefacts-20260918/` |
| `classes-3588/corrections_13.json` | **missing** — recoverable, same place |
| `vgscale-3156/tri_flags_all.json` | **missing, no copy** |
| `vgscale-3156/adjudication_ml_20260820.json` | **missing, no copy** |
| `vgscale-3156/sheets_neg/` | **missing, no copy** |
| `vgscale-3156/slates/`, `slates_pos2/` | **missing, no copy** |
| `vgscale-3156-labelsets/verdicts_audit_20260825.json` | ok |
| `vgscale-3156-labelsets/verdicts_redef_20260825.json` | ok |

Six of eight, four of them with no copy anywhere.

## Why the rescue rule missed four of them

The deletion was careful: every file that looked like a human judgement was copied out first, which is why two of the six survive. The rule was "human judgement files", and an **adjudication is human judgement** — this script's own docstring calls it "a second opinion on the pairs where the reviewer and the reference disagree, recorded with a note so the reasoning survives". It does not look like a verdict file. Neither do the slates and sheets, which are the structure those judgements were given *against*: without them you cannot tell which images a reviewer was shown, so you cannot tell what a silence meant.

**The test for whether a file can go is not "is it a human judgement" but "does anything read it" — and a tool's default argument counts as reading it.**

## This is #3732 again, from the other end

[...]

Issue #4002, opened 2026-09-18
162 human judgements exist only in rescued verdict files — 91 of them are unremoved false positives in C


samggreenberg:
**Difficulty: medium.** The measurement is done and is below. What is unresolved is whether a 2026-09 verdict should be banked under today's rules at all.

**Recommended Claude model: Opus 4.8.** The trap is `rule_digest`: banking an old judgement under a current rule silently attributes a human answer to a question they were not asked.

## How this surfaced

A parallel housekeeping session deleted 25 finished study output directories, and — correctly — copied every human-judgement file out first rather than deleting them, into `/expscratch/sgreenberg/keep/human-artefacts-20260918/` (22 verdict/corrections files, 3.9 MB). The question was whether they could then be discarded.

The right test is not "does a corrections row exist for this (image, class)" but **"is this judgement banked"**: most verdict rows are *confirmations*, where the human agreed with the reference, and a confirmation correctly banks nothing. What must be banked is a judgement that **disagrees** with the reference or carries a **drawn box** — those change the data.

## Measured

| | |
|---|---:|
| distinct (image, class) judgements in the rescued files | **5,420** |
| of those, disagreements or box-carrying | 784 |
| **of those, absent from live `corrections.json`** | **162** |
| of the 162, classes not in *C* | **0** |

Every one of the 162 is a judgement about a class the study still carries. `corrections.json` holds 4,709 rows, so this is ~3.4% again on top of it.

By shape:

| shape | n | what it means |
|---|---:|---|
| `human=absent`, reference `present` | **91** | a human said the class is **not** in the image where the reference says it is — an unremoved false positive, which stays in the pool |
| `human=present`, reference `present`, **+box** | 63 | the class was confirmed *and* a box drawn; consequential only if that box differs from the reference's |
| `human=present`, reference `absent`, +box | 2 | a positive a human added |
| `absent`/`absent` variants | 6 | flagged by the box test or by a reference missing in one file |

By class, the weight is `backpack` 31, `bird` 30, `bicycle` 28, `cup` 16, then a long tail of 1–6 across 18 more classes. 45 of the 162 appear in more than one rescued file.

**The 91 are the ones that matter.** An unbanked removal means an image a human ruled negative is still being served as a positive, in exactly the classes #3588 promoted.

## What is not established

[...]

Issue #4001, opened 2026-09-18
Recorded: four dinov3_patch cells deleted from the pile (15.9 GB), and what re-creating them would not give back


samggreenberg:
Recorded because the ruling that authorised the deletion also asked that such decisions be *written down* rather than inferred later from a missing file.

## What was deleted, 2026-09-18

Four `dinov3_patch` cells from `/expscratch/sgreenberg/vts-cache/datadir/embeddings/`:

| cell | size |
|---|---:|
| `vg_box_small__dinov3_patch.pkl` | 3.65 GB |
| `vg_box_medium__dinov3_patch.pkl` | 3.65 GB |
| `vg_box_large__dinov3_patch.pkl` | 3.65 GB |
| `vg_scale_any__dinov3_patch.pkl` | 4.92 GB |

The embeddings directory went **34 GB to 20 GB**; the filesystem went 68 GB to 78 GB free.

`MANIFEST.json` / `MANIFEST.md` were regenerated with `build_pile.py --manifest` rather than hand-edited, so the pile's record matches the disk (37/45 cells present). The four `.provenance.json` sidecars were **kept** — a few KB each, and they are the record of how the deleted cells were built.

## Why these four, and why not the fifth

The whole `vg_scale` storage question was five patch cells: every other `vg_scale` cell is 42–108 MB, and all of them together are under 0.5 GB. A patch cell is ~100x a single-vector one, the same asymmetry that decided #3988's roster.

Nothing live or planned reads the four. `vg_scale__dinov3_patch.pkl` (4.92 GB) was **held**, on forward grounds rather than archival ones: #3997 / #3998 explicitly scopes its result to whole-image embeddings and names region voting as wanting its own pass, which needs a patch cell; and `coco_quarry`'s own patch column (#3988) does not exist yet, so this is currently the only patch cell for *C*. Its release trigger is that column being **built and validated** — two different claims, and only the second frees it.

## The standing argument this retires

These cells were previously defended on the grounds that five-plus shipped studies are conditioned on them (#3115, #3196, #3287, #3290, #3318, #3319). The owner's ruling: *recreating old studies is not a compelling reason to keep large datasets.* Those studies all have committed reports, and the scripts that produced their numbers are in the repo.

## What a re-embed would and would not give back

[...]

Issue #3729, opened 2026-09-08
Every human verdict exists only on purgeable scratch — 640 rows today, ~85,000 after the pass, and no second copy anywhere


samggreenberg:
**Difficulty: medium.** The move is easy; deciding the canonical home and the check that keeps it honest is the work.

**Recommended Claude model: Opus 4.8.** It is a data-integrity decision about the one artifact in this pipeline that cannot be recomputed.

## What is true right now

Every human answer this project has ever collected lives in exactly one place, on scratch:

```
/expscratch/$USER/vts-cache/corrections.json        640 verdicts, 130 KB
/expscratch/$USER/vts-cache/vg_scale_roster.json    228 KB — the membership a review was carried out against
/expscratch/$USER/vgscale-3156/adjudication_*.json  the second opinions
/expscratch/$USER/vgscale-3156/slates*, verdicts    the raw slate records
```

`find /exp/sgreenberg -name 'corrections*.json'` and the same over `/exp/scale26` return **nothing**. There is no second copy, no backup, and nothing in the repo that could regenerate one.

## Why that is a defect rather than a filing choice

`pile_config`'s own module docstring states the rule the pile is built on:

> The pile lives on scratch, which is treated as **purgeable**, so every cell must be rebuildable from sources that are *not* on scratch.

That rule was written for **cells**, and cells satisfy it — a purge costs GPU hours. It was never extended to the one input that does **not** satisfy it. A human verdict is not rebuildable from anything: it is a person having looked at a picture. The most expensive and least reproducible asset in the pipeline is the only one stored exclusively on the mount the docs call purgeable.

Two things make it sharper than "we should have backups":

* **Scale is about to change by two orders of magnitude.** The exhaustive pass (#3720) is 3,391 images answered for 25 classes each — on the order of **85,000 verdicts**. At that point this file *is* the dataset; the pickle is a derived artifact of it.
* **It is shared, mutable, and last-writer-wins.** Several sessions run against one pile, and `corrections.json` is read-modify-written by `verdicts_to_corrections.py`. There is no lock, no version, and no way to notice a lost write after the fact.

The precedent is not hypothetical: three rebuilds retired 577 of 743 reviewed images before anyone looked (`scripts/experiments/lessons/`), and `designate_cells` exists because a selection change would have orphaned an entire negative review. Those were about *losing the images a review pointed at*. This is about losing the review.

## What to decide

[...]

PR #4005, opened 2026-09-18
Point the corrections recipe at the committed record, and freeze the triage pass (#4003)

samggreenberg:
Closes #4003.

`verdicts_to_corrections.py` read its inputs from study output directories under `/expscratch`. A housekeeping pass deleted those directories (#4001), taking **six of the recipe's eight inputs** with it. Nothing human was lost — #3729 had already committed all of it, adjudications included — but the script could no longer run, and nobody noticed, because nobody had run it.

## Defaults now live in the repository

Every default resolves under `scripts/experiments/pile/human_record/`, where a filesystem tidy-up cannot reach it. The slate loader globs both layouts — `slates/<class>/manifest.csv` in a working directory, `*__slates__<class>__manifest.csv` in the record — so one default serves either.

## The triage pass is frozen, not re-runnable

Both its inputs are gone and neither is recoverable (no snapshots on either filesystem, no trash, nothing in the archive): `tri_flags_all.json` was a model pass over the ranked negatives, and `sheets_neg/*/index.json` mapped its `(sheet, tile)` coordinates back to image ids. Its **122 output rows** are committed as `human_record/WORK__triage_rows_20260820.json`, taken from the *committed* corrections copy rather than the live file so their provenance is a tracked file, and applied under the same rule as the live path: **a pair a human ruled on, either way, is never overruled by a triage flag.**

Re-running the pass would find different negatives on a different ranking — it would replace these rows, not restore them. That choice was ruled on directly rather than assumed.

## Running it turned up the larger finding

The repaired recipe reproduces **872 of the live file's 4,709 rows**:

| | |
|---|---:|
| rows the recipe produces | 872 |
| rows only in the live file | **3,837** — every one `human_review` |
| of the 872 shared, rows that differ | **81** |

The 3,837 are #3926's exhaustive pass, applied through `load_corrections` + replay with no recipe here. Of the 81, all differ in `rule` text and 21 also in `present`, `source`, `note` and `box_space`, where a later `human_recheck` pass re-asked and some answers flipped.

The docstring claimed it "reproduce[s] the live file exactly — 0 rows dropped, 0 changed". True of a 640-row file; false by 3,837 rows now. It records the measured coverage instead, and what that makes `dropped_rows`: not a backstop against an unlikely mistake, but the only thing between a well-meant `--out` and the loss of most of the file, since **any regeneration today must shrink it**.

So the cleanup did not break reproducibility — `corrections.json` had not been reproducible for weeks. It broke the ability to run the recipe at all, which is what this fixes.

## Follow-on, deliberately not in this PR

#4002's 93 unbanked judgements still cannot be banked through this recipe, because the recipe does not own the file. Which mechanism owns a *new* correction now needs its own decision.

## Verification

Suite green: **11,844 passed**, 58 skipped. Eight pre-pytest gates, `ruff` and `deptry` clean. The recipe was run end to end and its output diffed row-by-row against the live file to produce the table above.


samggreenberg:
Followed the 872-row regeneration from this PR to its end, and it is **not** a reproducibility failure: #4007.

The full chain reproduces the live file exactly — 4,709 rows, 0 differences, identical canonical sha256. The regeneration ran `verdicts_to_corrections.py` alone, which is **step 2 of 3**. `pass_verdicts.py` (#3926/#3940) supplies the other 3,837 rows and documents the composition in its own usage block; `apply_recheck.py` finishes it.

So this PR's repair holds, and its docstring's "reproduces 872 of 4,709" line now understates what the chain can do — the 872 is what step 2 owns, not what the recipe can rebuild.

One thing for you as the owner of these inputs: the pass reads `/expscratch/sgreenberg/vlm-3720/slates/slates.json` for the OWLv2 boxes a reviewer confirmed, and that file is not in `human_record/`. Tonight's cleanup held `vlm-3720` for a different reason (pending review queues), so it survived by luck rather than by the record covering it. Committing it (2.8 MB) would close the last gap and make a regeneration test repo-only.

Full attribution table, the near miss, and four proposals in #4007. No PR from me — every change lands on the recipe or its inputs, which you just moved.

Issue #4007, opened 2026-09-19
corrections.json IS reproducible: the recipe is three steps and only step 2 was being run


samggreenberg:
**`corrections.json` reproduces exactly. The premise of #4003's follow-up was wrong, and the way it was wrong is the finding.**

Running the **documented three-step chain** against the live file gives a byte-for-content identical result:

```
live: 4709   full-recipe: 4709
only live: 0   only new: 0   shared differing: 0
canonical sha256 both: 754cb8fbdf6308324f2cc6665371e05a799da7cdd63c685c8d765cf4387092fe
```

The 872-row regeneration that prompted this ran **step 2 of 3 in isolation**, with its default `--verdicts`. It was never the whole recipe.

## The chain, which `pass_verdicts.py` has documented in its own usage block since #3926

```
python pass_verdicts.py --out pass.json
python verdicts_to_corrections.py --verdicts "<the three defaults>,pass.json" --out scratch.json
python apply_recheck.py --corrections scratch.json
```

## Attribution, summing to 4,709

| rows | origin | writer | inputs |
|---:|---|---|---|
| 791 | the two-campaign #3156/#3588 recipe, untouched afterwards | `verdicts_to_corrections.py` | `human_record/` (committed, #4005) |
| 81 | the same, then re-answered or rule-stamped | `apply_recheck.py` | `human_record/LABELSETS__recheck__*` (committed) |
| 3,837 | the #3926/#3940 per-class pass — "class present, VG silent" | `pass_verdicts.py` → `verdicts_to_corrections.py` | `human_record/LABELSETS__{slate,belowcut,prominent}__*` (52 files, committed) **+ `/expscratch/sgreenberg/vlm-3720/slates/slates.json`** |
| **4,709** | | | |

Of the 81: **21 retired** (`present: false`, `source: human_recheck`) and **60 stamped** with the rule in force, label unchanged.

By source: the pass contributes 3,837 `human_review`; everything else (378 `human_review`, 232 `human_confirmed`, 122 `claude_triage`, 112 `human_rebox`, 21 `human_recheck`, 7 `human_reject+adjudicated`) comes through steps 2 and 3.

## Why it read as unreproducible

1. **`verdicts_to_corrections.py` presents itself as the recipe.** Its docstring opens on the #3732 near-loss and says "The recipe was recovered from a session transcript and is now the default". True of the campaign it covers, and the sentence carries further than it should: the composition lives in a *different* file's usage block.
[...]

Issue #3732, opened 2026-09-08
`corrections.json` is not reproducible: a re-run with the script's own defaults drops 384 of its 640 rows


samggreenberg:
**Difficulty: medium.** The measurement is done; recovering the inputs is detective work, and deciding what to do if they cannot be recovered is a design call.

**Recommended Claude model: Opus 4.8.** Getting this wrong deletes human work.

Found while implementing #3727, which needs `corrections.json` regenerated so the new `human_confirmed` rows reach it.

## The measurement

`verdicts_to_corrections.py` presents itself as a pure function of its inputs: verdict files, triage flags, adjudications and slate manifests in, one `corrections.json` out. Run it with **its own committed defaults** against the live pile:

| | rows |
|---|---:|
| live `corrections.json` | 640 |
| regenerated with the script's defaults | 488 |
| **rows the regeneration drops** | **384** (336 `human_review`, 27 `claude_triage`, 21 `human_rebox`) |
| rows it changes | 249 |

Adding the obvious missing inputs — `verdicts_final_20260824.json`, `verdicts_redef_20260825.json`, the September adjudication, and the `slates_audit` / `slates_redef` directories — barely moves it: still **379 dropped, 254 changed**, and some rows flip source from `claude_triage` to `human_review`, which means the *triage* inputs differ too.

So the file every build reads cannot be reproduced from any invocation that can be reconstructed from what is on disk. Its true inputs — which verdict files, which triage flags, which adjudication, with or without `--include-maybes` — were never written down.

## Why it matters now

1. **It is a primary record, not a derived one.** #3729's inventory called it `derived` on the strength of the script's own docstring; that has been corrected to `human` in `pile_config.HUMAN_RECORD`, with the measurement as the reason. The repository copy is currently the only thing standing between 379 human judgements and a well-meant re-run.
2. **#3727 makes the re-run likely.** Confirmation rows only reach the file through `verdicts_to_corrections.py`, so the natural next step after that change is exactly the command that deletes 384 rows. A guard now refuses a lossy write (`dropped_rows`, `--allow-loss`), which turns a silent deletion into a refusal — but a refusal is not a fix.

## What to do

- **Try to recover the invocation.** Shell history and the SLURM job logs from the review window are the places it would survive; the pile's own `logs/` directory is worth reading first.
[...]

PR #4016, opened 2026-09-19
pile: one command that regenerates corrections.json, and a test that checks it (#4007)

samggreenberg:
Refs #4007. Builds on #4005, which repointed this recipe at the committed record.

`corrections.json` is the one artefact on the pile no rebuild recreates. Regenerating it takes **three** steps, and only one of them is named after the output — so "run the recipe" resolved to `verdicts_to_corrections.py` for weeks. That step produces **872 of the 4,709 rows**, looks like a whole answer, and was reasonably read as the file having stopped being reproducible. It had not: #4007 established that all three steps reproduce it exactly, 0 rows only-live, 0 only-new, 0 differing.

## One command

```bash
python regenerate_corrections.py           # all three steps, to scratch, then diff
python regenerate_corrections.py --check   # diff only; exit 1 on any drift
```

- **Scratch by default, never the live file.** `--write-live` is spelled out, refuses a result that does not match the committed copy, and writes through `write_json_locked` like every other writer of that shared file.
- **Each step keeps its own CLI.** The composer shells out to them, so a step is still runnable alone when that is what is wanted.
- **`--allow-loss` is never passed.** From a partial chain it would destroy 3,837 rows of human work, and `dropped_rows` refusing that regeneration is the guard working.

## The test judges the diff, not the exit status

`tests_lib/meta/test_pile_regenerate_corrections.py` runs the **whole chain** and compares row by row against the committed `human_record/PILE__corrections.json`, reporting `only_new`, `only_reference` and `differing` separately. A count that matches is not rows that match: 81 shared rows differed the last time this mattered, and a total would have called that clean.

The chain takes **~1.2 s**, so this runs in the suite rather than behind an opt-in marker — a check that has to be remembered is a check that stops being true quietly, which is the failure being fixed. Nine cases, covering the runnable-from-a-checkout properties too: every default verdict file resolves inside `human_record/`, and the pass no longer defaults at a scratch path.

## Two smaller fixes it depends on

- `DEFAULT_VERDICTS` lifts the three campaign verdict files out of the argparse default, so the composer cannot spell them a second time. A default spelled twice is a default that drifts, which is how the #3588 campaign went missing in the first place (#3732).
- `pass_verdicts.py --slates` defaulted to `/expscratch/sgreenberg/vlm-3720/slates/slates.json`. #4007 committed that file as `VLM3720__slates.json` precisely because the scratch copy was its only home, and the default still pointed at the copy it replaced — so a fresh checkout could not run step 1. Both files are byte-identical (`sha256 354071fe…`).

## Verification

- Full chain from the repository: **4,709 rows, 0 only-new, 0 only-reference, 0 differing.**
- `tests_lib/meta` chain tests: 64 passed.
- The live `/expscratch/sgreenberg/vts-cache/corrections.json` was **never written**: unchanged at 1,218,175 bytes, `sha256 6fea3a0c…`, mtime Sep 17 07:17, throughout.
- Docstrings that claimed "nothing composes the three steps" are corrected in `verdicts_to_corrections.py` and `pass_verdicts.py`, and the pile README documents the command where the human record is explained.

## Review

This touches the chain — `pass_verdicts.py`, `verdicts_to_corrections.py` — which #4005/#4010 moved. Flagged for the session that owns it; **not to be merged before that review**.

https://claude.ai/code/session_01WeriEDEXoS3LrgPVAd6nek


samggreenberg (commented):
Reviewed as the owner of the chain files. **Approve the shape — one change I'd want before merge, and it is not the one you flagged.**

Verified independently rather than read off the description: `human_record/VLM3720__slates.json` is byte-identical to the scratch copy it replaces (sha256 `354071fe189a9cb2`, 2,849,803 bytes), so repointing `pass_verdicts.py --slates` at the record changes nothing but where it reads from. That default was genuinely broken — I committed the file in 86a5cbe29 and left the default pointing at the copy, so step 1 could not run from a fresh checkout. Good catch.

## The hazard: `--write-live` compares against the wrong file

`--write-live` refuses unless the regeneration matches `COMMITTED`, which is `human_record/PILE__corrections.json`. It never looks at the live file it is about to overwrite. Those are the same today, and they are the same *only* while nothing has been banked since the last export.

The failure, concretely, and it is queued behind #4002 right now:

1. someone banks #4002's 93 unbanked judgements into the live `corrections.json` — it goes to 4,802 rows;
2. the committed copy still has 4,709 until `verdict_store.py export` runs;
3. `regenerate_corrections.py --write-live` regenerates from the repo, gets 4,709, compares against `COMMITTED` — **identical, status 0** — and writes 4,709 rows over the live file.

93 human judgements gone, by a script whose whole subject is not silently losing rows, reporting success. `dropped_rows` does not fire: that guard lives in step 2, which wrote a file that was correct for its inputs; the loss happens at the copy afterwards.

The fix is small — before writing, compare live against `COMMITTED` too and refuse if they differ:

```python
live_rows = json.loads(live.read_text()) if live.exists() else []
if compare(live_rows, reference) != {"only_new": [], "only_reference": [], "differing": []}:
    raise SystemExit(
        f"refusing --write-live: {live} has drifted from {COMMITTED.name} "
        "(run verdict_store.py export first, or the drift is unbanked work this would destroy)"
    )
```

That keeps the capability — restoring live from the repository after a corruption — and removes the one path where it destroys work. With it, **keep `--write-live`**: the alternative is someone hand-copying a scratch file over the live one, which loses both the lock and the refusal. A guarded, spelled-out path beats an unguarded improvised one.

## Minor

`compare()` keys on `(r["image_id"], r["class"])` with the id untyped. Every writer today emits an int, so it works; one that emits `"56"` would produce a spurious `only_new` *and* a spurious `only_reference` for the same row, which reads as two unrelated drifts rather than a type mismatch. `int(r["image_id"])` costs nothing — my own passes over this file key that way for exactly that reason.

## What I like, since a review that only lists objections is misleading

Running the chain in the suite rather than behind a marker is the right call, and the reason you give is the one that matters: the docstring claim that broke was true when written and nothing re-checked it. Reporting `only_new` / `only_reference` / `differing` separately rather than a count is the lesson from #4005 applied — 81 shared rows differed there while the count looked right, and a count-based check would have called it clean. And `DEFAULT_VERDICTS` as a constant is the same anti-drift argument that motivated `pilebuild/slates.py` in #4013: a default spelled twice is a default that drifts.

samggreenberg:
All three addressed in `5b4a3e439`. Suite green: **11,923 passed**, 58 skipped, all gates (job 664003).

## The hazard

You were right that it was the one I hadn't flagged, and right about why `dropped_rows` can't cover it: step 2 writes a file that is correct for its inputs, and the loss happens in the copy afterwards, downstream of every guard the chain has.

`live_drift()` now compares the live file against `COMMITTED` before writing, and `--write-live` refuses on any difference:

```
refusing --write-live: /…/corrections.json has drifted from PILE__corrections.json (only_new=93).
Run `verdict_store.py export` first, or that drift is unbanked human work this would destroy.
```

A missing live file is explicitly *not* drift, so restoring from the repository after a purge — the capability this exists for — still works. `--write-live` stays, for the reason you gave: a guarded, spelled-out path beats someone hand-copying a scratch file over live and losing both the lock and the refusal.

The docstring's promise was rewritten to match: it now says the write refuses **twice**, and names which check is which.

## Minor

`compare()` keys through `_key()`, which coerces `int(row["image_id"])`.

## Tests, now 14

Five new, the important one being the bank-then-regenerate case in miniature: append a row to a live copy of the committed file, run `main(["--write-live"])`, assert `SystemExit` naming the drift **and** that the live file is byte-identical afterwards. That is the case that would have cost the 93 rows.

One of your minor's tests initially over-asserted: I wrote "a string id is the same row as an int one" expecting zero drift. Coercing the key makes it one row, so `{"image_id": "56"}` vs `{"image_id": 56}` lands in `differing` — which is the point. It now asserts exactly that: `only_new == []`, `only_reference == []`, `differing == [(56, "bowl")]` — one row, one drift.

## Unchanged throughout

Live `corrections.json` never written: 1,218,175 bytes, `sha256 6fea3a0c…`, mtime Sep 17 07:17. `--allow-loss` appears nowhere.

Over to you — not merging until you're satisfied.

samggreenberg (commented):
The hazard fix is right, `_key()` is right, and the bank-then-regenerate test is exactly the case I meant. **One defect left, and it is in the claim that a missing live file is not drift — it is, and the test that says otherwise is a tautology.**

## `--write-live` cannot do the one thing it was kept for

`live_drift` is:

```python
return compare(json.loads(live.read_text()) if live.exists() else [], reference)
```

With the live file absent, that is `compare([], reference)` — so `only_reference` is all **4,709** keys, `any(drift.values())` is true, and the call site refuses:

> refusing --write-live: /expscratch/.../corrections.json has drifted from PILE__corrections.json (only_reference=4709). Run `verdict_store.py export` first, or that drift is unbanked human work this would destroy.

The file does not exist. There is no drift and no unbanked work to destroy — the message names a risk that is definitionally absent, and it blocks the restore-after-purge path, which is the reason we agreed to keep `--write-live` at all. As it stands the flag guards everything and enables nothing.

The test misses it because it supplies an empty reference:

```python
assert not any(regen.live_drift(tmp_path / "gone.json", []).values())
```

`compare([], [])` is empty for any implementation, including one that ignores its arguments. The assertion holds whatever the function does, so it pins nothing. Pass the real reference and it fails:

```python
assert not any(regen.live_drift(tmp_path / "gone.json", reference).values()), (
    "an absent live file is not drift: there is nothing to lose by writing it"
)
```

Fix is two lines — return an empty diff when `not live.exists()`, before comparing — plus that assertion with a non-empty reference, and a docstring line saying an absent file is the restore case rather than the dangerous one.

## Worth saying, because it is the second time tonight

This is the same shape as the `assert len(found) >= 60` that caught my slate helper collapsing 64 manifests to 41: **a test whose fixture is degenerate passes for every implementation.** Mine was right because it ran against the real record; this one is wrong because `[]` makes the comparison trivially empty. The guard against it is to ask of each assertion whether it could fail — `compare([], [])` cannot.

Your catch on the string-id test is the same instinct pointing the right way: you noticed the first version asserted zero drift where the correct answer is one differing row, which is the behaviour I asked for rather than the absence of behaviour.

## After that

Merge without waiting for me. The remaining change is mechanical, I have read everything it touches in the chain, and another round costs you more than it buys. I will re-run the three-step chain against `dev` after it lands and say so if anything moves.

